Issue #43, opened 2022-02-22
Evaluate quality of reports against oss-fuzz projects


oliverchang:
The reports for the OSS-Fuzz projects where the introspector successfully runs are now public at:

https://oss-fuzz-introspector.storage.googleapis.com/

We should go and evaluate these to ensure their quality, before we make them more broadly available to users. 

e.g. I browsed a few and found some issues/weirdness:

- [arduinojson](https://oss-fuzz-introspector.storage.googleapis.com/arduinojson/inspector-report/20220221/fuzz_report.html) has weird fuzzer names (e.g. "srcarduinojsonextrasfuzzingjson_fuzzer.cpp")

- [libxml2](https://oss-fuzz-introspector.storage.googleapis.com/libxml2/inspector-report/20220221/fuzz_report.html#Fuzzer:-xml.c) has no functions hit data for `xml.c`.

Navidem:
These projects fail to generate `fuzz_report.html` report while the build is successful:

```
abseil-cpp
alembic
bitcoin-core
boringssl
casync
cel-cpp
circl
clamav
dart
dav1d
double-conversion
draco
fribidi
grpc-httpjson-transcoding
harfbuzz
http-pattern-matcher
jbig2dec
libavc
libcacard
libcoap
libhevc
libjxl
librawspeed
libspectre
libssh
libtpms
lwan
open62541
opencensus-cpp
openexr
opensc
ots
perfetto
pffft
s2opc
spice-usbredir
systemd
tarantool
tcmalloc
tidy-html5
tink
tmux
tpm2-tss
upb
```

A quick check in couple of logs shows: `INFO:fuzz_data_loader: - found 0 profiles to load`

Navidem:
> A quick check in couple of logs shows: INFO:fuzz_data_loader: - found 0 profiles to load

For the ones encountering `0 profiles to load` this can be the culprit:
introspector is storing `fuzzerLogFile*.data` and `fuzzerLogFile*.data.yaml` in /work while [compile script](https://github.com/google/oss-fuzz/blob/9bf510557ca863d1de8d9fedb1664316b0bea5ae/infra/base-images/base-builder/compile#L210) expects them in /src

This is the case for `lwan`.

Navidem:
> These projects fail to generate `fuzz_report.html` report while the build is successful:

#52 helped to reduce this number to 19 projects:

```
abseil-cpp
bitcoin-core
cel-cpp
circl
clamav
dart
[...]

PR #52, opened 2022-03-01
Prefix all profile log files

Navidem:
Fuzzer can be compiled in different directories. We have to make sure the profiles are created in a pre-determined location to let compile script be able to find and use them.

Fixes issues like [this](https://github.com/ossf/fuzz-introspector/issues/43#issuecomment-1051258838)

### DIFF
--- a/llvm/lib/Transforms/FuzzIntrospector/FuzzIntrospector.cpp
+++ b/llvm/lib/Transforms/FuzzIntrospector/FuzzIntrospector.cpp
@@ -336,8 +336,12 @@ std::string FuzzIntrospector::getNextLogFile() {
   std::string TargetLogName;
   std::string RandomStr = GenRandom(10);
   int Idx = 0;
+  std::string prefix = "";
+  if (getenv("SRC")) {
+    prefix = std::string(getenv("SRC")) + "/";
+  }
   do {
-    TargetLogName = formatv("fuzzerLogFile-{0}-{1}.data", std::to_string(Idx++), RandomStr);
+    TargetLogName = formatv("{0}fuzzerLogFile-{1}-{2}.data", prefix, std::to_string(Idx++), RandomStr);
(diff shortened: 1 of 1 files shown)


llvm/lib/Transforms/FuzzIntrospector/FuzzIntrospector.cpp
```diff
@@ -336,8 +336,12 @@ std::string FuzzIntrospector::getNextLogFile() {
   std::string TargetLogName;
   std::string RandomStr = GenRandom(10);
   int Idx = 0;
+  std::string prefix = "";
+  if (getenv("SRC")) {
```
DavidKorczynski, line 340: Should we make the environment variable more explicit to fuzz introspector, i.e. not `SRC` which is an OSS-Fuzz name? I added two environment variables recently: https://github.com/ossf/fuzz-introspector/blob/50d02ba2094524bcda679605c541b2a456d039bf/llvm/lib/Transforms/FuzzIntrospector/FuzzIntrospector.cpp#L251 and https://github.com/ossf/fuzz-introspector/blob/50d02ba2094524bcda679605c541b2a456d039bf/llvm/lib/Transforms/FuzzIntrospector/FuzzIntrospector.cpp#L855  I think it would be nice to prefix the environment variables in the same way. I see that I mixed it up slightly (`FUZZ_INTROSPECTOR` and `FUZZINTRO_CONSTANTS`). How about prefixing all environment variables with `FUZZINTRO_`, and then perhaps changing `SRC` to `DST` (or perhaps `OUTDIR` or something else)?

Navidem: yes, it make sense to unify environment variable names.

DavidKorczynski (approved):

Issue #45, opened 2022-02-22
Improve default sort order of tables


oliverchang:
Current tables seem to always be sorted on the first column, which can be a little confusing in some cases. On e.g. https://oss-fuzz-introspector.storage.googleapis.com/libxml2/inspector-report/20220221/fuzz_report.html#functions_cov_hit_1, it may be better to order by "source code lines" descending by default rather than alphabetically.

PR #54, opened 2022-03-02
Sort table by source code lines

AdamKorcz:
If a table has the column "source code lines" it will sort by that table descending.

Closes https://github.com/ossf/fuzz-introspector/issues/45

### DIFF
--- a/post-processing/styling/custom.js
+++ b/post-processing/styling/custom.js
@@ -15,8 +15,16 @@ $( document ).ready(function() {
 //    vary)
 function createTables() {
   $.each(tableIds, function(index, value) {
+    createTable(value, false);
+  });
 
-    // Get number of rows in this table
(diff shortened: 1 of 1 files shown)


DavidKorczynski:
Very nice! 

We can get this in now, but, would it be possible to make it more general such that we can indicate which column is the sortable one when creating the HTML table in the python code, for example as an argument to this function:

https://github.com/ossf/fuzz-introspector/blob/8a8e3b7f6a4229f86ac97a0e28ae2b7eaeecc98d/post-processing/fuzz_html.py#L112

AdamKorcz:
> We can get this in now, but, would it be possible to make it more general such that we can indicate which column is the sortable one when creating the HTML table in the python code, for example as an argument to this function:

Sure, will look into that.

DavidKorczynski (approved):

Issue #116, opened 2022-03-17
allow users to specify a "threat model"


DavidKorczynski:
Allow users to specify a "threat model" which means specifying which function/files/paths should be included/excluded in the analysis

Ref: 
https://github.com/ossf/fuzz-introspector/issues/35

https://github.com/ossf/fuzz-introspector/issues/107

https://github.com/ossf/fuzz-introspector/issues/74#issuecomment-1071241152

DavidKorczynski:
For now this can be done with some form of command line options to fuzz introspector - like a config file.

DavidKorczynski:
This is now done.

Issue #67, opened 2022-03-09
unify oss-fuzz integrations


DavidKorczynski:
Currently we have two oss-fuzz integrations: the integration upstream in oss-fuzz and the one local in fuzz-introspecctor. Ideally the one in OSS-Fuzz upstream should also be used locally here in fuzz-introspector. The only thing we really need from fuzz-introspector's perspective is some scripting to make it easy testing new developments locally

DavidKorczynski:
@Navidem @oliverchang -- we now use the upstream OSS-Fuzz fuzz-introspector integration here as well. We still need a small bit of patching to support testing during development, so the main things we do in the patching is just adding support for copying in fuzz-introspector from the file system rather than cloning it from github.

Two additional changes that are currently needed in the patching :
- removing `--coverage-url` from the arguments to the post-processing: https://github.com/ossf/fuzz-introspector/blob/96b7aee16a40e54b788fecd5520de555a69211a7/oss_fuzz_integration/oss-fuzz-patches.diff#L21-L22
- avoiding launch of http server: https://github.com/ossf/fuzz-introspector/blob/96b7aee16a40e54b788fecd5520de555a69211a7/oss_fuzz_integration/oss-fuzz-patches.diff#L57-L58

will update the docs accordingly!

DavidKorczynski:
> I don't know if OSS-Fuzz pulls the fuzz-introspector images locally when downloading the latest images?

Before I update the docs it might be nice to clarify this -- can users run fuzz-introspector locally using OSS-Fuzz without having to build images locally? When I run the following commands `base-clang` without introspector is used, which means the fuzz-introspector clang is not used and thus no results:
```
sudo docker system prune -a
git clone https://github.com/google/oss-fuzz
cd oss-fuzz
python3 infra/helper.py builder_fuzzers --sanitizer=introspector htslib
```

DavidKorczynski:
> > I don't know if OSS-Fuzz pulls the fuzz-introspector images locally when downloading the latest images?
> 
> Before I update the docs it might be nice to clarify this -- can users run fuzz-introspector locally using OSS-Fuzz without having to build images locally? When I run the following commands `base-clang` without introspector is used, which means the fuzz-introspector clang is not used and thus no results:
> 
> ```
> sudo docker system prune -a
> git clone https://github.com/google/oss-fuzz
> cd oss-fuzz
> python3 infra/helper.py builder_fuzzers --sanitizer=introspector htslib
> ```
> 
[...]

PR #123, opened 2022-03-17
oss-fuzz: use default fuzz-introspector config

DavidKorczynski:
Ref: https://github.com/ossf/fuzz-introspector/issues/116


oliverchang:
Drive by comment/question: Do we still need these patches in these repo when most of the changes have been upstreamed into OSS-Fuzz?

DavidKorczynski:
In short, I am not 100% sure, but likely no. We have an issue open for unifying: https://github.com/ossf/fuzz-introspector/issues/67

The main point is that it's nice to have some mechanism for easy patching here as we also develop on fuzz-introspector, e.g. if we add new environment flags etc. and need to test it.

Issue #131, opened 2022-03-18
calltree issues in libdwarf


DavidKorczynski:
Consider the following piece of calltree from libdwarf:

![Screenshot from 2022-03-18 16-46-58](https://user-images.githubusercontent.com/657617/159046560-85ce4aad-0263-415d-9d74-9e3e3cafdb66.png)

The calltree shows as if calltree index 45 and 46 are in the same funciton and right after each other, but in practice index 46 is actually within a completely different function. I don't know if some inlining is happening that we don't account for, or something different. But its' definitely off:

https://github.com/davea42/libdwarf-code/blob/ba03f108e88a7b53014ed1f78a7667039224bb4b/src/lib/libdwarf/dwarf_object_detector.c#L724-L866

DavidKorczynski:
This is to do with inlining. I think for now we should enforce a compiler flag to forces no inlining when compiling for fuzz-introspector.
![Screenshot from 2022-03-18 21-12-03](https://user-images.githubusercontent.com/657617/159083753-64bfb4da-d3b3-458f-8ffd-16472245045c.png)

PR #136, opened 2022-03-18
oss-fuzz: force no inlining

DavidKorczynski:
Ref: https://github.com/ossf/fuzz-introspector/issues/131

CC @Navidem

### DIFF
--- a/oss_fuzz_integration/oss-fuzz-patches.diff
+++ b/oss_fuzz_integration/oss-fuzz-patches.diff
@@ -6,7 +6,7 @@ index 475e5a7d..fca8165d 100644
  
  ENV SANITIZER_FLAGS_thread "-fsanitize=thread"
  
-+ENV SANITIZER_FLAGS_instrumentor "-flto"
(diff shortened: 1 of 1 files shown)


oliverchang:
Thanks for fixing this!

@Navidem could you please help with incorporating this in oss-fuzz as well.

Going forward, would you be able to just patch oss-fuzz instead? @DavidKorczynski I completely understand the usefulness of quickly iterating and testing changes with patches, but it would reduce a lot of duplicated work, reduce turnaround and avoid  missed patches for our production deployment :)

DavidKorczynski:
> Going forward, would you be able to just patch oss-fuzz instead? @DavidKorczynski I completely understand the usefulness of quickly iterating and testing changes with patches, but it would reduce a lot of duplicated work, reduce turnaround and avoid missed patches for our production deployment :)

yes definitely -- I wasn't quite sure if I should or not but will help ensuring sync on the oss-fuzz side moving forward as well.

DavidKorczynski:
This was unintentionally reverted by https://github.com/ossf/fuzz-introspector/pull/141

PR #142, opened 2022-03-19
oss-fuzz: cleanup integration

DavidKorczynski:


DavidKorczynski:
ref https://github.com/ossf/fuzz-introspector/issues/67

PR #146, opened 2022-03-21
oss-fuzz: restore no inlining

DavidKorczynski:
Ref: https://github.com/ossf/fuzz-introspector/pull/136#issuecomment-1074030590


DavidKorczynski:
Actually this shouldn't be merged -- it should just be verified that it works. The merge should only happen on the OSS-Fuzz side.

PR #160, opened 2022-03-30
Use per-target coverage report in fuzzer details

Navidem:
With coverage reports by target produced on OSS-Fuzz side, we can use it in "Fuzzer details" section.
It makes sense for the project overview report to link to overall coverage, so I did not change it.

### DIFF
--- a/post-processing/fuzz_html.py
+++ b/post-processing/fuzz_html.py
@@ -766,6 +766,14 @@ def create_calltree(
     write_wrapped_html_file(calltree_html_string, calltree_html_file, blocker_idxs)
     return calltree_html_file
 
+def get_target_coverage_url(coverage_url: str, target_name: str) -> str:
+    """
+    This function changes overall coverage URL to per-target coverage URL. Like:
+        https://storage.googleapis.com/oss-fuzz-coverage/<project>/reports/<report-date>/linux
+        to
+        https://storage.googleapis.com/oss-fuzz-coverage/<project>/reports-by-target/<report-date>/<target-name>/linux
+    """
+    return coverage_url.replace("reports", "reports-by-target").replace("linux", f"{target_name}/linux")
 
 def create_fuzzer_detailed_section(
         profile: fuzz_data_loader.FuzzerProfile,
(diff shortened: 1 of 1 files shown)


DavidKorczynski:
one issue here is that it won't work on local runs, which is not ideal. How about setting some guard up that recognises if we are in OSS-Fuzz, and, if so, will use OSS-Fuzz urls?

DavidKorczynski (changes requested):
one issue here is that it won't work on local runs, which is not ideal. How about setting some guard up that recognises if we are in OSS-Fuzz, and, if so, will use OSS-Fuzz urls?

Navidem:
> How about setting some guard up that recognises if we are in OSS-Fuzz

Yes, this is actionable. Why local runs do not have per-target reports, is it because of using an older OSS-Fuzz version?

DavidKorczynski:
> Why local runs do not have per-target reports, is it because of using an older OSS-Fuzz version?

No, this is because local runs don't rely on the same infrastructure for running the fuzzers and storing coverage reports. `https://storage.googleapis.com/oss-fuzz-coverage/<project>/reports-by-target/<report-date>/<target-name>/linux` this links is not used by local runs. 

This script https://github.com/ossf/fuzz-introspector/blob/main/oss_fuzz_integration/run_both.sh is used for local runs

DavidKorczynski:
I think the smartest is to set up a guard that checks if it's an oss-fuzz environment, and if not, won't use the per coverage reports. I can look into how we can make it work locally better down the line - but this seems the smartest for getting to v1.

Navidem:
> This script https://github.com/ossf/fuzz-introspector/blob/main/oss_fuzz_integration/run_both.sh is used for local runs

This still is using infra/helper.py which stores the coverage reports in `out/report/linux`. That's why I think if we bump OSS-Fuzz on local instances, we can get `reports-by-target`.

This can be left for later, as I see it requires the patching process updated.

DavidKorczynski:
ah fair enough - let's merge this then and take it from there. I can run locally afterwards and fix up if something is broken

DavidKorczynski (approved):

Issue #155, opened 2022-03-29
Arrow inspector reports fail to load in Chrome


oliverchang:
e.g. https://oss-fuzz-introspector.storage.googleapis.com/arrow/inspector-report/20220328/fuzz_report.html#Complex-functions-with-low-coverage gives me a "Aw, Snap!" after a while. Likely we're hitting OOM limits.

DavidKorczynski:
The page loads now https://oss-fuzz-introspector.storage.googleapis.com/arrow/inspector-report/20220523/fuzz_report.html but fails to load proper in that I can't get data into the functions tables. The amount of data is huge: 76K functions and 420K cyclomatic complexity
![Screenshot from 2022-05-28 18-01-31](https://user-images.githubusercontent.com/657617/170835355-0a7c88e0-fbc3-41e3-afc2-0a097b48fae8.png)

The page was enabled loading by pushing some of the loading into javascript rather than in the python server (https://github.com/ossf/fuzz-introspector/pull/163) However, we should probably try to push this further such that the functions table can be loaded.

Some efforts that may be worth exploring here
- automatically exclude more content from the report. This can potentially be done more aggressively if the project is large.
- doing some partial loading of functions tables, although this will have undesired side effects
- partial loading controlled by the backend server -- am not sure I think this is something we should start to be doing though.

AdamKorcz:
I will look into optimizing the UX further for large projects like Arrow.

PR #163, opened 2022-03-31
Load table data from frontend for 2 tables

AdamKorcz:
This creates a json object for the "Project functions overview" table and the "All functions overview".

The rows a of the tables are then added to the table in the frontend instead of the backend.

The goal is to improve page load speed and avoid issues like: https://github.com/ossf/fuzz-introspector/issues/155. Locally the report page loads much faster for me. The two tables still take a couple of seconds to add the data, but the loading is now isolated to the tables instead of blocking the whole report page. If this becomes a problem for large projects, the UX can be further improved by adding a loader for each table.

There is a bit of refactoring to do as a result of this PR, but it would be good to test this out a bit before going too far down this way of improving page load speed. The refactoring would be to ensure all html generation of these two tables is removed from the backend.

@DavidKorczynski PTAL

### DIFF
--- a/post-processing/styling/custom.js
+++ b/post-processing/styling/custom.js
@@ -8,16 +8,16 @@ $( document ).ready(function() {
           var elemId;
           elemId = $(this).closest("a").attr('id');
           $(".left-sidebar-content-box > div > a").each(function( index ) {
-              console.log("link: ", $(this).attr("href").replace("#", ""));
-              if($(this).attr("href").replace("#", "")===elemId) {
-                if(!$(this).hasClass("activeMenuText")) {
-                  $(this).addClass("activeMenuText");
-                }
-              } else {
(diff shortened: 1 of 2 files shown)


post-processing/styling/custom.js
```diff
@@ -8,20 +8,28 @@ $( document ).ready(function() {
...
+                $(this).removeClass("activeMenuText");
+              }
+            }            
           })
         };
       });
     });
+
+  console.log(all_functions_table_data.length);
+  for(var i=0;i<all_functions_table_data.length;i++) {
```
DavidKorczynski, line 27: Needed?

DavidKorczynski (changes requested):
LGTM, but maybe remove the loop I commented?

DavidKorczynski (approved):

Issue #111, opened 2022-03-17
Improve html generation


AdamKorcz:
As fuzz-introspector adds more features, managing the html from the backend becomes increasingly complex.

Now in the backend we generate both the data for the html as well as the html itself. A better way going forward could be to only generate the data in the backend and place it in the html files as json objects. When a user loads a page, we could generate the reports in the frontend based on the data.

This separates the data-generation from the report-generation.

PR #185, opened 2022-04-04
Improve report page load speed

AdamKorcz:
Improves the report page load speed by:

- Loading all rows once in the "Project functions overview" and "All functions overview" tables instead line by line. 
- Loading the data in the "Functions hit (dynamic analysis based)" tables in the frontend instead of from `fuzz_html.py` as suggested in https://github.com/ossf/fuzz-introspector/issues/111


post-processing/fuzz_html.py
```diff
@@ -1236,6 +1278,7 @@ def create_html_report(
     # Section with details about each fuzzer, including calltree.
     #############################################
     logger.info(" - Creating section with details about each fuzzer")
+    fuzzer_table_data = typing.Dict[str, Any]
```
DavidKorczynski, line 1281: This should be `fuzzer_table_data: typing.Dict[str, Any] = dict()`

DavidKorczynski: Fixed in https://github.com/ossf/fuzz-introspector/commit/b4f6e9d373a87460bfb2637c4dabb140644662b4

PR #191, opened 2022-04-07
post-process: refine fuzz-blocker definition

DavidKorczynski:
Remove the requirement of fuzz blockers not blocking "higher up code".
The problem with it is that sometimes a blocker will be in the middle of
a sequence of red nodes and this is fairly counter-intuitive. Having
looked at more reports now it seems to make more sense to simply focus
on sequences of red nodes and just identifying blockers by looking for
transition points.


DavidKorczynski:
still testing this

DavidKorczynski:
This is in my experience a lot nicer -- merging

PR #198, opened 2022-04-11
Added Dockerfile

marcinguy:
Hi,

PR request to add Dockerfile (adding also GCC 10 and CLANG/LLVM 12 - not needed)

Helps to use it.

Thanks,

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,20 @@
 FROM ubuntu:20.04 AS fuzz-introspector
 
-#GCC-10 and CLANG/LLVM 12
+# GCC-10 and CLANG/LLVM 12
+# Copyright 2021 Fuzz Introspector Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+# 
+#     http://www.apache.org/licenses/LICENSE-2.0
+# 
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# 
 
 
 ARG DEBIAN_FRONTEND=noninteractive
--- a/Dockerfile
+++ b/Dockerfile
@@ -47,20 +47,11 @@ ENV LLVM_CONFIG=llvm-config-12
 RUN apt update
 RUN apt-get install python3.8-venv -y
 RUN apt-get install texinfo cmake -y
-WORKDIR /
-RUN git clone https://github.com/ossf/fuzz-introspector
-WORKDIR /fuzz-introspector
-
-# Get python dependencies
-RUN python3 -m venv .venv
-RUN . .venv/bin/activate && pip3 install -r requirements.txt
-
-# Build custom clang with Fuzz introspector LLVM pass
-RUN ./build_all.sh
-
-
 
 RUN echo 'alias joe="jupp --wordwrap"' >> ~/.bashrc
 RUN echo 'export PS1="[fuzz-introspector]$PS1"' >> ~/.bashrc
 ENV IS_DOCKER="1"
+
+WORKDIR /src
+ENTRYPOINT "/src/build_all.sh"
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,66 @@
+FROM ubuntu:20.04 AS fuzz-introspector
+
+#GCC-10 and CLANG/LLVM 12
+
+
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,66 @@
+FROM ubuntu:20.04 AS fuzz-introspector
```
DavidKorczynski, line 1: Could you add the proper license?

Dockerfile
```diff
@@ -0,0 +1,66 @@
...
+RUN update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-10 0
+RUN update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-10 0
+
+ENV LLVM_CONFIG=llvm-config-12
+
+RUN apt update
+RUN apt-get install python3.8-venv -y
+RUN apt-get install texinfo cmake -y
+WORKDIR /
+RUN git clone https://github.com/ossf/fuzz-introspector
```
DavidKorczynski, line 51: Instead of cloning from github, could you do it so it uses the local source?

DavidKorczynski (changes requested):
Thanks a lot! A few comments for now

Dockerfile
marcinguy: @DavidKorczynski Added. Or it should be a different one?

Dockerfile
marcinguy: Changed to work with local source.

Dockerfile
DavidKorczynski: looks good but please update it to 2022

marcinguy: Done

DavidKorczynski (changes requested):
Does this Dockerfile install a lot more packages than is actually needed? I would prefer if it doesn't install a lot of stuff that is not needed to run fuzz-introspector.

I have one locally as follows:
```
FROM ubuntu:20.04

ENV DEBIAN_FRONTEND noninteractive
RUN apt-get update && \
    apt-get upgrade -y && \
    apt-get install -y libc6-dev binutils && \
    apt-get autoremove -y

RUN apt-get update && \
    apt-get install -y software-properties-common && \
    add-apt-repository ppa:git-core/ppa && \
    apt-get update && \
    apt-get install -y \
        binutils-dev \
        build-essential \
        curl \
        git \
        jq \
        autoconf bison texinfo cmake

RUN apt-get install -y \
    python3 \
    python3-pip \
    python3.8-venv \
    clang-12

COPY . /fuzz-introspector
WORKDIR /fuzz-introspector

# First install python virtual environment
RUN python3 -m venv .venv && \
    . .venv/bin/activate && \
    pip install -r requirements.txt && \
    ./build_all.sh
```
Although it's not tested fully so am unsure about the completeness of it, however, it has much fewer packages installed

WDYT?

marcinguy:
Both are fine to me.

Took a time to build yours. It builds.

However, it happens to be bigger than mine, I think also installs more packages:

fuzz-introspector  - my Dockerfile
fuzz-introspector2 - your paste  

```
REPOSITORY                               TAG          IMAGE ID       CREATED         SIZE
fuzz-introspector2                       Dockerfile   8fae482a728e   6 minutes ago   85GB
fuzz-introspector                        Dockerfile   05e8331a6515   2 days ago      83.6GB
```
Maybe you can add:

`--no-install-suggests --no-install-recommends`

Which should make it smaller & faster.

Don't have time to test it however.

Any Dockerfile would help people, since before it wasn't there.

Also I would not copy the source to image/container (`COPY . /fuzz-introspector`). I would mount the local dir and execute container with the mount.

Great work on fuzz-introspector.

Hope that helps.

marcinguy:
Well you will have the folder in the Docker and it will work on it. When Docker exists/stops your results are outside of the container. AFAIK You might don't even need to start container to use binaries in the folder (depending on linkage etc)

In your setup you need to start container and enter it. Somehow feed your project there (via mount or SSH copy etc).

Whatever works :) Not an expert on Docker.

DavidKorczynski:
sounds good, thanks! Am not sure I understand the reasoning about packages though, is your thought that mine installs more packages because of the size? My reasoning came from your Dockerfile having a lot of packages installed by way of `apt`. I think perhaps the size difference comes from me copying in more, e.g. my `oss-fuzz` folder is pretty sizeable within the `oss_fuzz_integration`.

In either case let's get this merged and we'll take it from there! Thanks again!

DavidKorczynski (approved):

marcinguy:
Thanks. I can try your Dockerfile again, but add `--no-install-suggests --no-install-recommends` to `apt` and see if it still works. Since I think your does install suggested and recommended packages by default. As I understand they're typically not required. I think I saw many other packages being added that way. 

This could be the most optimal sizewise. 

I get back as I know.

PR #208, opened 2022-04-15
post-process: modularise fuzz driver synthesis analyis

DavidKorczynski:


DavidKorczynski:
This separates the synthesis part out of the optimal target analysis. It also shows how a plugin (synthesis) can use another plugin (optimal targets).

PR #237, opened 2022-04-26
post-process: extract focus function to html report

DavidKorczynski:


oliverchang:
Thanks for this!

I'm a little worried about the amount of info going into the human readable html report though. This information seems more useful to automated fuzzing infra instead as a JSON file and can be omitted from the human readable report?

oliverchang:
@DavidKorczynski ping :)

DavidKorczynski:
Sounds good! I will fix this up tomorrow (Thursday) UK time. I will also set it up such that the analyses run without displaying it on the HTML report (also added this https://github.com/ossf/fuzz-introspector/issues/239 to remove some of the bloat from the report).

DavidKorczynski:
> Thanks for this!
> 
> I'm a little worried about the amount of info going into the human readable html report though. This information seems more useful to automated fuzzing infra instead as a JSON file and can be omitted from the human readable report?

https://github.com/ossf/fuzz-introspector/pull/245 removes the HTML from the main page and writes it to a JSON file instead. Please see that PR for details.

PR #240, opened 2022-04-27
calltree: restructure html

AdamKorcz:
Moves the expand/collapse symbol to the left side.


DavidKorczynski:
Cool! Would it be possible to get also get the call tree indices on the left side?

Issue #239, opened 2022-04-27
refine all function table


DavidKorczynski:
The function table is great for displaying a lot of data about each function in the module. However, it can feel a bit bloated and is easy to get lost in the data. It would be nice to refactor it such that on the main report we have a trimmed down version of it, and then we have another page with all details if necessary. Columns that could be removed:
- bb count (cyclomatic complexity shows related info)
- instr count (cyclomatic complexity shows related info)
- Fuzzers runtime hit (because Func lines hit % shows this)

Other columns?

DavidKorczynski:
Fixed by https://github.com/ossf/fuzz-introspector/pull/247

The solution also allows users to interactively pick and choose which columns to show. Thanks @AdamKorcz

PR #246, opened 2022-04-28
html report: Improve styling

AdamKorcz:
Adds color styling of percentages in two tables.

Adds a button to specify visibility of columns in tables.


DavidKorczynski:
Could we also modify the default columns shown? i.e. https://github.com/ossf/fuzz-introspector/issues/239

Could we also disable "Arg types" by default? I feel given that it's easy to enable columns we should prefer to have a cleanly-designed table that is easy to understand at first sight, and then users can decide which columns to show based on their needs, WDYT?

Issue #249, opened 2022-04-28
c++ coverage issue


DavidKorczynski:
There is an issue where callsites in c++ code is not recognised in the calltree, an example is jsoncpp:
![Screenshot from 2022-04-28 21-17-16](https://user-images.githubusercontent.com/657617/165838469-e7644da5-e924-44af-a2be-8e2f48e95282.png)

The call site for Message() is hit, however, it is not shown in the calltree.

DavidKorczynski:
FYI this is different -- and more minor -- to https://github.com/google/oss-fuzz/issues/7635

PR #250, opened 2022-04-28
post-process: dont normalise coverage function

DavidKorczynski:
Function names are not normalised when put in the coverage map. They are
in fact demangled.

Fixes: https://github.com/ossf/fuzz-introspector/issues/249


DavidKorczynski:
Once this one is merged then we should bump on OSS-Fuzz. Will just test this through

AdamKorcz (approved):

Issue #41, opened 2022-02-16
Improve testing


DavidKorczynski:
We currently have some facilities to test fuzz-introspector, including:

- Testing local benchmarks: https://github.com/ossf/fuzz-introspector/blob/main/tests/run_simple_tests.sh
- Testing OSS-Fuzz integrations by ensuring fuzz-introspector runs to completion for a given project: https://github.com/ossf/fuzz-introspector/blob/main/oss_fuzz_integration/test_projects.sh

There are room for improvements in terms of testing. This is an umbrella issue for testing improvements that commits can refer to.

DavidKorczynski:
Setting this to higher priority since we need to provide better guards for OSS-Fuzz bumps:
https://github.com/google/oss-fuzz/issues/7635

oliverchang:
Thanks! Another idea for OSS-Fuzz integration may be to pin to a "stable" release for OSS-fuzz, and we can bump after manually verifying that a few reports on the latest revision looks OK. 

May not be needed if the tests are comprehensive enough though.

DavidKorczynski:
Ref https://github.com/ossf/fuzz-introspector/pull/287

DavidKorczynski:
Removed priority and enhancement labels as the testing infrastructure is now at a much better place than when this issue was created.

Keeping the issue open so it can be referenced in commits.

PR #251, opened 2022-04-30
oss-fuzz: add proper testing set up

DavidKorczynski:
Add testing set up for oss-fuzz that can automatically analyse
properties against a set of projects. This is achieved by making
fuzz-introspector write .json files with data about the run which is
then processed by the testing framework.

The testing framework has some hard-coded boundary checks on the data
for a given project. It has to be boundaries rather than fixed values
since the data may change from run to run depending on the fuzzing
results.

In order to make the check complete it needs to have hard-coded values
from a diverse set of projects, in particular both c and c++ projects.

Another benefit of this PR is that the output in json format can be consumed by other tools.

Ref: https://github.com/ossf/fuzz-introspector/issues/41


DavidKorczynski:
todo: add a few more projects to test against and add documentation on how to run it

Navidem (approved):
LGTM, thanks!

AdamKorcz (approved):

PR #252, opened 2022-05-10
post-process: output multiple focus functions per fuzzer

DavidKorczynski:
Now multiple focus functions are output. Example file:
```
{
    "fuzzers": {
        "hts_open_fuzzer": {
            "focus-functions": [
                "view_sam",
                "bcf_hdr_read",
                "bcf_hdr_init",
                "cram_read_file_def",
                "possibly_expand_bam_data",
                "sam_parse1",
                "bgzf_read_block",
                "sam_hdr_destroy",
                "sam_read1"
            ]
        }
    }
}
```

The bulletpoints from https://github.com/ossf/fuzz-introspector/pull/245 still hold. Note `focus-function` was renamed to `focus-functions` in the json.

### DIFF
--- a/post-processing/analyses/fuzz_engine_input.py
+++ b/post-processing/analyses/fuzz_engine_input.py
@@ -156,7 +156,8 @@ class FuzzEngineInputAnalysis(fuzz_analysis.AnalysisInterface):
         )
 
         html_string += (
-            f"<p>Use one of these functions as input to libfuzzer with flag: -focus_function name </p>"
+            f"<p>Use one of these functions as input to libfuzzer with flag: "
+            f"-focus_function name </p>"
             f"<pre><code class='language-clike'>"
             f"-focus_function={focus_functions}"
             f"</code></pre><br>"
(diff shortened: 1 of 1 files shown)


DavidKorczynski:
CC @oliverchang

Navidem:
Are we making sure the func names are demangled?  https://github.com/ossf/fuzz-introspector/pull/245 shows a case that is not.

DavidKorczynski:
We do not, and the reason for this is that libFuzzer will need the mangled version (as far as I could tell) -- it's declared here in the fuzz driver https://github.com/llvm/llvm-project/blob/ce583b14b2ec37b1c168bb92020680cb452502b3/compiler-rt/lib/fuzzer/FuzzerDriver.cpp#L760

Navidem:
> We do not, and the reason for this is that libFuzzer will need the mangled version (as far as I could tell) -- it's declared here in the fuzz driver https://github.com/llvm/llvm-project/blob/ce583b14b2ec37b1c168bb92020680cb452502b3/compiler-rt/lib/fuzzer/FuzzerDriver.cpp#L760

Well, I had a different experience at the time trying it for xpdf:
This did not work: `-focus_function=_ZN11JBIG2Stream12readSegmentsEv`
It only worked when I used `-focus_function="JBIG2Stream::readSegments()"`

As a matter of the fact, I just tried it again and can confirm that mangled name does not work: it fails with this message:
```
ERROR: Failed to set focus function. Make sure the function name is valid (_ZN11JBIG2Stream12readSegmentsEv) and symbolization is enabled.
```

DavidKorczynski:
Ah perfect, thanks for verifying! let's demangle then!

Issue #262, opened 2022-05-13
statically type the project


DavidKorczynski:
Make the project more stable with static typing throughout the codebase. 

This would have fixed an erroneous string concatenation https://github.com/ossf/fuzz-introspector/pull/261

DavidKorczynski:
Ideally we should be able to run `mypy` with `--strict`

DavidKorczynski:
This has progressed and most of it is statically typed now.

PR #264, opened 2022-05-13
post-process: add static typing to lot of codebase

DavidKorczynski:
The goal is to enable better prediction of bugs and make the code more
readable.

Ref: https://github.com/ossf/fuzz-introspector/issues/262


DavidKorczynski:
This snowballed quite a bit: once I added some typing the type checker would identify missing types or even issues. Will test this through.

AdamKorcz (approved):

PR #266, opened 2022-05-13
calltree analysis: statically type

DavidKorczynski:
main point is to statically type `get_fuzz_blockers` but the PR includes some styling too.


DavidKorczynski:
and the typing finds a potential (over-approximated and not actually viable though) issue :)!

DavidKorczynski:
Looking deeper, the issue highlighted by mypy may actually be a true bug, so all in all good with the typing!

AdamKorcz (approved):

PR #269, opened 2022-05-18
Add Branch profile and branch-level blocker detection

Navidem:
This PR adds these:

- Branch profiling at LLVM plugin: identify branch sides and their corresponding cyclomatic complexity
- Loading branch profile in fuzz_data_loader.py
- Loading branch coverage information in fuzz_cov_load.py
- Making the calls to branch-level blocker detection in fuzz_analysis.py

### DIFF
--- a/llvm/lib/Transforms/FuzzIntrospector/FuzzIntrospector.cpp
+++ b/llvm/lib/Transforms/FuzzIntrospector/FuzzIntrospector.cpp
@@ -1304,7 +1304,7 @@ std::string FuzzIntrospector::getInsnDebugInfo(Instruction *I) {
   if (Loc != NULL)
   {
     ret_string = Loc->getFilename().str() +
-              +":" + std::to_string(Loc->getLine());
+              +":" + std::to_string(Loc->getLine()) + "," + std::to_string(Loc->getColumn());
   }
   else {
     logPrintf(L1, "No debug info!!\n");
--- a/llvm/lib/Transforms/FuzzIntrospector/FuzzIntrospector.cpp
+++ b/llvm/lib/Transforms/FuzzIntrospector/FuzzIntrospector.cpp
@@ -113,6 +113,23 @@ typedef struct FuzzerModuleIntrospection {
       : FuzzerFileName(A), AllFunctions(B) {}
 } FuzzerModuleIntrospection;
 
+typedef struct BranchSidesComplexity {
+  std::string TrueSideString;
+  size_t TrueSideComp;
+  std::string FalseSideString;
+  size_t FalseSideComp;
+
+  BranchSidesComplexity ()
+      : TrueSideString(), TrueSideComp(0), FalseSideString(), FalseSideComp(0) {}
+  BranchSidesComplexity(std::string TS, size_t TC, std::string FS, size_t FC)
+      : TrueSideString(TS), TrueSideComp(TC), FalseSideString(FS), FalseSideComp(FC) {}
+} BranchSidesComplexity;
+
+typedef struct BranchProfileEntry {
+  std::string BranchString;
+  BranchSidesComplexity BranchSidesComp;
+} BranchProfileEntry;
+
 // YAML mappings for outputting the typedefs above
 template <> struct yaml::MappingTraits<FuzzerFunctionWrapper> {
   static void mapping(IO &io, FuzzerFunctionWrapper &Func) {
@@ -156,6 +173,25 @@ template <> struct yaml::MappingTraits<FuzzerModuleIntrospection> {
     io.mapRequired("All functions", introspectorModule.AllFunctions);
   }
 };
+
+template <> struct yaml::MappingTraits<BranchSidesComplexity> {
+  static void mapping(IO &io, BranchSidesComplexity &branchSidesComp) {
+    io.mapRequired("TrueSide", branchSidesComp.TrueSideString);
+    io.mapRequired("TrueSideComp", branchSidesComp.TrueSideComp);
+    io.mapRequired("FalseSide", branchSidesComp.FalseSideString);
+    io.mapRequired("FalseSideComp", branchSidesComp.FalseSideComp);
+  }
+};
+
+template <> struct yaml::MappingTraits<BranchProfileEntry> {
+  static void mapping(IO &io, BranchProfileEntry &bpe) {
+    io.mapRequired("Branch String", bpe.BranchString);
+    io.mapRequired("Branch Sides", bpe.BranchSidesComp);
+  }
+};
+LLVM_YAML_IS_SEQUENCE_VECTOR(BranchProfileEntry)
+
+
 // end of YAML mappings
 
 namespace {
@@ -175,6 +211,8 @@ typedef struct CalltreeNode {
 } CalltreeNode;
 
 static FILE *OutputFile = stderr;
+
+std::map<const Function *, size_t> FuncComplexityMap;
 
 struct FuzzIntrospector : public ModulePass {
   static char ID;
@@ -220,6 +258,15 @@ struct FuzzIntrospector : public ModulePass {
   void logPrintf(int LogLevel, const char *Fmt, ...);
   bool runOnModule(Module &M) override;
 
+  void branchProfiler(Module &M);
+  SmallPtrSet<BasicBlock *, 32> findReachables(BasicBlock *src);
+  std::pair<size_t, size_t> findComplexities(SmallPtrSet<BasicBlock *, 32>, SmallPtrSet<BasicBlock *, 32>, std::map<BasicBlock *, size_t>);
+  std::string getInsnDebugInfo(Instruction *I);
+  std::string getBBDebugInfo(BasicBlock *BB);
+  // void writeOutMap(std::map<std::string, std::vector<std::pair<std::string, size_t>>>, std::string);
+  // void writeOutMap(std::map<std::string, BranchSidesComplexity> outMap, std::string fileName);
+  void writeOutMap(std::vector<BranchProfileEntry> outMap, std::string fileName);
+  size_t calculateBBComplexity(BasicBlock *BB);
 };
 } // end of anonymous namespace
 
@@ -348,6 +395,8 @@ bool FuzzIntrospector::runOnModule(Module &M) {
   // Log data about all functions in the module
   std::string nextYamlName = nextCalltreeFile + ".yaml";
   extractAllFunctionDetailsToYaml(nextYamlName, M);
+
+  branchProfiler(M);
 
   logPrintf(L1, "Finished introspector module\n");
   return true;
@@ -1037,6 +1086,8 @@ FuzzerFunctionWrapper FuzzIntrospector::wrapFunction(Function *F) {
     FuncWrap.CyclomaticComplexity = 1;
   }
 
+  FuncComplexityMap[F] = FuncWrap.CyclomaticComplexity;
+
   std::set<StringRef> FuncReaches;
   std::vector<CalltreeNode *> Nodes;
   // TODO: extractCalltree should not be run on all functions in this manner.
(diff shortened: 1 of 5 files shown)


DavidKorczynski:
Am at Kubecon but will review this tomorrow night!

Navidem:
> Am at Kubecon but will review this tomorrow night!

No worries! Good luck with your presentations.

llvm/lib/Transforms/FuzzIntrospector/FuzzIntrospector.cpp
```diff
@@ -1132,3 +1184,234 @@ static RegisterStandardPasses
         PM.add(new FuzzIntrospector());
       });
 */
+
+void FuzzIntrospector::branchProfiler(Module &M) {
+
+  SmallPtrSet<const BasicBlock *, 32> visistedBBs;
+  SmallVector<const BasicBlock *, 32> worklist;
+  std::string outFileName = getNextLogFile() + ".branchProfile.yaml";
+  std::vector<BranchProfileEntry> outMap;
```
DavidKorczynski, line 1193: I think in general we should stick to LLVM naming convention and have the first letter upper case

DavidKorczynski: The same applies to the other parts of the diff

llvm/lib/Transforms/FuzzIntrospector/FuzzIntrospector.cpp
```diff
@@ -1132,3 +1184,234 @@ static RegisterStandardPasses
...
+
+  return allReachables;
+}
+
+// Calculate complexities reachable from each reachable unique BBs
+// TODO: for now, this function just accounts for complexity of callee functions from 
+// the reachable BBs. We can do better by carefully calculating complexity of the reachable
+// regions of the code i.e. intra-procedural complexity by counting the number of unique 
+// edges and nodes for each reachable set.
+std::pair<size_t, size_t> FuzzIntrospector::findComplexities(SmallPtrSet<BasicBlock *, 32> trueReachable, SmallPtrSet<BasicBlock *, 32> falseReachable, std::map<BasicBlock *, size_t> BBComplexityMap) {
```
DavidKorczynski, line 1299: Please run `clang-format` on the code -- I think this line is a bit too long for `clang-formats` style

llvm/lib/Transforms/FuzzIntrospector/FuzzIntrospector.cpp
```diff
@@ -1132,3 +1184,234 @@ static RegisterStandardPasses
...
+  //   result = getInsnDebugInfo(&I);
+  //   if (result.first.length() > 0) {
+  //     break;
+  //   }
+  // }
+  return result;
+}
+
+// void FuzzIntrospector::writeOutMap(std::map<std::string, std::vector<std::pair<std::string, size_t>>> outMap, std::string fileName) {
+// void FuzzIntrospector::writeOutMap(std::map<std::string, BranchSidesComplexity> outMap, std::string fileName) {
```
DavidKorczynski, line 1380: Do we want to keep these two lines? I'm okay with leaving comments like this in it helps your dev process, but if they are not needed for future work I suggest removing

llvm/lib/Transforms/FuzzIntrospector/FuzzIntrospector.cpp
```diff
@@ -1132,3 +1184,234 @@ static RegisterStandardPasses
...
+
+  auto YamlStream = std::make_unique<raw_fd_ostream>(
+      fileName, EC, llvm::sys::fs::OpenFlags::OF_None);
+  yaml::Output YamlOut(*YamlStream);
+
+  YamlOut << outMap;
+}
+
+// Return cyclomatic complexity of called functions in the BB
+size_t FuzzIntrospector::calculateBBComplexity(BasicBlock *BB) {
```
DavidKorczynski, line 1393: In the context of this method, do we need it? I'm thinking from the perspective of accummulating the cyclomatic complexity of an arbitrary set of functions is really easy in the post-processing as we have all the data there. It might be nicer to keep the LLVM module a bit more lean and, instead of accumulating cyclomatic complexity only output the functions callable by a given basic block?

Perhaps even more than that we might extract all of the callees for each basic block and then output this? I have a feeling this sort of data might be useful for other post-processing tasks?

In general, could we decrease some of the logic in the compiler plugin and do more of the reasoning in the post-processing? Down the line this might also be nice for example for doing the same technique for other languages?

I think the same principle applies to `findComplexities` -- it does a bunch of computation and then uses that as data in the data collector, but could we somehow push the computation out of the compiler plugin and into the post-processing part, and then use the compiler plugin as more of a data extractor?

llvm/lib/Transforms/FuzzIntrospector/FuzzIntrospector.cpp
```diff
@@ -1132,3 +1184,234 @@ static RegisterStandardPasses
...
+      auto BI = dyn_cast<BranchInst>(TI);
+      if (BI && BI->isConditional()) {
+        auto side0 = BI->getSuccessor(0); 
+        auto side1 = BI->getSuccessor(1);
+        auto BILoc = BI->getDebugLoc();
+
+        auto reachable0 = findReachables(side0);
+        auto reachable1 = findReachables(side1);
+
+        std::pair<size_t, size_t> complexities = findComplexities(reachable0, reachable1, BBComplexityMap);
```
DavidKorczynski, line 1220: Please see below where I have some questions about `findComplexities`.

DavidKorczynski (changes requested):
Looks cool!

Overall LGTM -- I think my main point is that we should reduce the amount of computation in the compiler plugin and push more towards making the compiler plugin a data extractor and then the post-processing for essentially all computation on the data. The benefit of this is both for dev time (it's much faster working with the post-processing), the broadness (that other's might want to use the raw data too) and also to prepare for when we eventually add new languages/compilers as then we really want all the computation in the post-processing to unify analyses in one place. WDYT?

llvm/lib/Transforms/FuzzIntrospector/FuzzIntrospector.cpp
DavidKorczynski: This is now above :)

DavidKorczynski:
Notice about the above review -- if this PR is blocking https://github.com/google/oss-fuzz/pull/7709 then I'm happy to address the things over iterations if it helps get the ball moving faster

llvm/lib/Transforms/FuzzIntrospector/FuzzIntrospector.cpp
Navidem: Thanks for the comments. Actually once I have these implemented and tested locally I started to come up with different approaches to profile branches (please check the latest changes in the design doc I shared with you). In a nutshell I totally agree with delegating computation to post-process, and probably I will deprecated these logic from LLVM side and only extract the list of functions called by each branch side. This is the approach that I am more inclined with as providing more raw data gives such more flexibility on the types of analysis we can perform later in post-process.

llvm/lib/Transforms/FuzzIntrospector/FuzzIntrospector.cpp
Navidem: Generally agreed, and addressed in prev comment :)

Navidem:
> Looks cool!
> 
> Overall LGTM -- I think my main point is that we should reduce the amount of computation in the compiler plugin and push more towards making the compiler plugin a data extractor and then the post-processing for essentially all computation on the data. The benefit of this is both for dev time (it's much faster working with the post-processing), the broadness (that other's might want to use the raw data too) and also to prepare for when we eventually add new languages/compilers as then we really want all the computation in the post-processing to unify analyses in one place. WDYT?

Thanks for looking into this, I agree with the general idea of delegating most of the computation to post-processing. The computations here in the LLVM pass are not finalized and I have changes in mind which I think we can make in iterations.

llvm/lib/Transforms/FuzzIntrospector/FuzzIntrospector.cpp
Navidem: Done, wow it made a lot of changes!

llvm/lib/Transforms/FuzzIntrospector/FuzzIntrospector.cpp
Navidem: removed.

oliverchang:
> Notice about the above review -- if this PR is blocking [google/oss-fuzz#7709](https://github.com/google/oss-fuzz/pull/7709) then I'm happy to address the things over iterations if it helps get the ball moving faster

That would be great, thanks! This is an experimental feature that we're going to iterate on. We just need to make sure this doesn't accidentally break any of the existing analysis.

llvm/lib/Transforms/FuzzIntrospector/FuzzIntrospector.cpp
DavidKorczynski: > please check the latest changes in the design doc I shared with you

Will get this done tomorrow morning -- apologies for the delay here.

llvm/lib/Transforms/FuzzIntrospector/FuzzIntrospector.cpp
DavidKorczynski: Ah, I see -- quick question, did you try to rebuild it after the clang-format? I'm asking because sometimes the `#includes` will be re-arranged in a manner that breaks the build. It's quite annoying -- in either case I will rebuild, test and merge tomorrow but if you already know then let me know!

Navidem: Yes, I was curious about the many changes almost all over the cpp file, so tried to re-build the clang image and it finished successfully.

llvm/lib/Transforms/FuzzIntrospector/FuzzIntrospector.cpp
```diff
@@ -1122,13 +1200,239 @@ PreservedAnalyses FuzzIntrospectorPass::run(Module &M, ModuleAnalysisManager &AM
...
+        auto Side1Comp = Complexities.second;
+
+        std::pair<std::string, std::string> DbgExtracts;
+        DbgExtracts = getInsnDebugInfo((Instruction *)BI);
+        auto BRstring = DbgExtracts.first;
+        if (BRstring.length() == 0) {
+          continue; // Failed to get debug info
+        }
+        DbgExtracts = getBBDebugInfo(Side0, BILoc);
+        auto Side0String = DbgExtracts.first;
```
DavidKorczynski, line 1256: In general for my liking there's a bit too much use of `auto`. The problem I have is when `auto` is used when the type is not explicitly clear from reading the code, e.g. in return values from functions that are not ubiquitous or fields in classes/structs that are not ubiquitous -- and I don't think we should use `auto` and then move the "String" to the variable name.

post-processing/fuzz_data_loader.py
```diff
@@ -87,6 +87,40 @@ def __init__(self, elem) -> None:
         self.total_cyclomatic_complexity: int = 0
 
 
+class BranchProfile:
+    """
+    Class for storing information about conditional branches collected by LLVM pass.
+    """
+    def __init__(self) -> None:
+        self.branch_line = str()
```
DavidKorczynski, line 95: perhaps `branch_pos` instead? point being it's more than just the line number, which I thought it was at first given the `line` name

post-processing/fuzz_data_loader.py
```diff
@@ -629,3 +665,31 @@ def load_input_bugs(bug_file: str) -> List[InputBug]:
             continue
 
     return input_bugs
+
+
+def read_branch_data_file_to_profile(filename: str, BP_dict: dict):
```
DavidKorczynski, line 670: I think we tend to have lower-case variable/param names in the Python code

post-processing/fuzz_analysis.py
```diff
@@ -47,6 +49,21 @@ def analysis_func(self,
         pass
 
 
+class BlockedSide(Enum):
+    TRUE = 1
+    FALSE = 2
+
+
+class FuzzBlocker:
```
DavidKorczynski, line 57: I would probably call this `FuzzBranchBlocker` or so

post-processing/fuzz_analysis.py
```diff
@@ -234,6 +252,10 @@ def get_hit_count_color(hit_count):
         n1.cov_forward_reds = forward_red
         n1.cov_largest_blocked_func = largest_blocked_name
 
+    branch_blockers = detect_branch_level_blockers(profile, branch_profiles)
+    logger.info(f"[+] found {len(branch_blockers)} branch blockers.")
+    # TODO: use these results appropriately ...
```
DavidKorczynski, line 257: Am curious, is the goal to use this based on the logging for now? I'm asking because I didn't see the explicit use of the stuff for now

DavidKorczynski (approved):

llvm/lib/Transforms/FuzzIntrospector/FuzzIntrospector.cpp
Navidem: Fair point. I plan to refactor this part of the code.

PR #279, opened 2022-05-26
Branch blocker: Extract reached funcs for each branch side

Navidem:
This changes the way branches are profiled:
- for each conditional branch, identifies reachable functions for each branch side
- the post-process measures the complexity of un-hit reachable functions for each side

### DIFF
--- a/llvm/lib/Transforms/FuzzIntrospector/FuzzIntrospector.cpp
+++ b/llvm/lib/Transforms/FuzzIntrospector/FuzzIntrospector.cpp
@@ -1444,6 +1444,7 @@ FuzzIntrospector::getBBDebugInfo(BasicBlock *BB, DILocation *PrevLoc) {
   Instruction *CurrTI, *CurrI;
   DILocation *CurrLoc;
 
+  // Traverse all dummy BBs associated with the previous Loc.
   do {
     CurrTI = CurrBB->getTerminator();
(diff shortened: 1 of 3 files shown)


llvm/lib/Transforms/FuzzIntrospector/FuzzIntrospector.cpp
```diff
@@ -1317,6 +1335,52 @@ FuzzIntrospector::findReachables(BasicBlock *Src) {
   return AllReachables;
 }
 
+vector<StringRef> FuzzIntrospector::findReachableFuncs(BasicBlock *Src) {
```
DavidKorczynski, line 1338: Could you add a comment about the function? In particular that this is local to a function and not program-wide reachability?

DavidKorczynski (commented):
I think we should start to move the data extracted for branch profiling into the `wrapFunction` function to follow the structure of the existing logic. We can do this by having `BranchProfileEntry` as an element in `FuzzerFunctionWrapper`, which this should make things follow neatly into the existing code

Navidem:
> I think we should start to move the data extracted for branch profiling into the `wrapFunction` function to follow the structure of the existing logic. We can do this by having `BranchProfileEntry` as an element in `FuzzerFunctionWrapper`, which this should make things follow neatly into the existing code

That makes sense, done!

Issue #280, opened 2022-05-27
Python target integration


DavidKorczynski:
Integrate support for Python fuzzing.

Roughly, the steps:
1) Add a set of test cases such that once these are analysed properly we have basic support for python
2) Add static analysis part to cover the basic data needed by the post-processing. In this context use a (modified) version of https://github.com/vitsalis/PyCG
3) Add support for extracting the static data on a project-wide level
4) Demonstrate basic capabilities of fuzz-introspector excluding runtime coverage information
5) Add runtime coverage features, using the coverage approach from OSS-Fuzz
6) Demonstrate full end-to-end features
7) Cleanup/refactor to ensure datastructures neatly fit the existing fuzz-introspector implementation

DavidKorczynski:
The current state is:
![Screenshot from 2022-07-29 22-36-48](https://user-images.githubusercontent.com/657617/181846335-62b51582-5e93-491b-9b84-5586c42f5ce8.png)


This is from `pyyaml` OSS-Fuzz project run by way of https://github.com/ossf/fuzz-introspector/blob/main/oss_fuzz_integration/run_both.sh

DavidKorczynski:
Example using it on `idna`:
![Screenshot from 2022-10-26 04-54-51](https://user-images.githubusercontent.com/657617/197930966-79ff7b8d-af81-4e18-8fee-9bd9e5464b9e.png)

DavidKorczynski:
`glom`:


![Screenshot from 2022-10-26 13-14-33](https://user-images.githubusercontent.com/657617/198023544-3c83dd75-fcfd-4c64-9209-489d247b6edc.png)

DavidKorczynski:
Python is now working in general. Closing this and will make more specific issues for further Python work.

PR #295, opened 2022-05-30
post-processing: enable reading of Python output

DavidKorczynski:
This provides a few minor changes and some bug fixes, which enables
reading of the Python frontend's output as is. There will have to be
further changes to make the project more friendly for having data from
sources in different languages.

Ref: https://github.com/ossf/fuzz-introspector/issues/280


DavidKorczynski:
Screenshot of Python calltree in the HTML report

Screenshot from 2022-05-30 17-31-42

PR #309, opened 2022-06-08
post-processing: add link to repository issues

DavidKorczynski:
When fuzz-introspector is used on OSS-Fuzz it may be non-obvious where
to file suggestions and issues. To guide the user to the
fuzz-introspector repository include a short message in the navigation
bar directing where to file this.

### DIFF
--- a/post-processing/fuzz_html_helpers.py
+++ b/post-processing/fuzz_html_helpers.py
@@ -82,8 +82,13 @@ def html_get_navbar(title: str) -> str:
             </g>
         </svg>
     </div>
-    <div class="top-navbar-title">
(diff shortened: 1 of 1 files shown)


DavidKorczynski:
@AdamKorcz @Navidem WDYT?

AdamKorcz:
The URL will not be a link now. Should we wrap it in an `<a>` tag?

DavidKorczynski:
> Good idea! The URL will not be a link now. Should we wrap it in an `<a>` tag?

I think we should!

Navidem (approved):
LGTM, thanks!

PR #311, opened 2022-06-08
frontend: python: enable scanning of additional libraries

DavidKorczynski:
This makes it easier to include more files in the AST generation.

### DIFF
--- a/frontends/python/main.py
+++ b/frontends/python/main.py
@@ -32,8 +32,20 @@ def main():
         help="Package containing the code to be analyzed",
         default=None
     )
+    parser.add_argument(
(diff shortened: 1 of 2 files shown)


DavidKorczynski:
Python frontend is progressing, this is the generated calltree for https://github.com/ossf/fuzz-introspector/blob/5317f46e0a77f06f606d3e834a9120a40f66fc66/tests/python/test4/fuzz_test.py

Screenshot from 2022-06-08 21-51-23

DavidKorczynski:
Ref https://github.com/ossf/fuzz-introspector/issues/280

PR #318, opened 2022-06-10
Add multicore support.  Requires python library psutil

swirsz:


DavidKorczynski (approved):
Awesome, in a follow-up PR could you add the ability to make this optional/configurable? It would be nice to be able to control it when using `run_both.sh` in the same way it's possible to control the amounts of seconds to fuzz.

DavidKorczynski:
I just realised this has implications on https://github.com/ossf/fuzz-introspector/blob/main/oss_fuzz_integration/test_projects.sh as we now execute more than the 10 seconds cpu time which causes the expected results to be different. This can be fixed if we make it optional/configurable how many cpus to use.

Issue #302, opened 2022-06-06
Support multiple cores in runtime analysis


swirsz:
When running locally, large projects would benefit from libfuzzer being called with multiple cores.  To test this, I modified run_fuzzer in get_full_coverage.py with "-workers=n -jobs=n" and used psutil.cpu_count().

capstone 80: (default setup)
real	19m7.849s
user	0m1.883s
sys	    0m5.875s
LOC 48.26% (126880/262901)

capstone 800: (default setup)
real	33m1.631s
user	0m2.590s
sys	    0m16.393s
LOC 88.64% (268790/303248)

capstone 80: (32 cores)
real	19m37.386s
user	0m3.986s
sys	    0m8.049s
LOC 90.62% (274815/303248)

LOC obtained from corpus/inspector-report/covreport/linux/report.html

DavidKorczynski:
This is fair and I think we should add this.

However, perhaps nicer is to enable the use of existing corpus rather than running the fuzzer to generate corpus.

DavidKorczynski:
Would you be able to assist her with the implementation?

What we need to do is enable `get_full_coverage.py` to pass arbitrary commands to `run_fuzzer` here:
https://github.com/ossf/fuzz-introspector/blob/7493818cc99dd6ca85d81e81d0cda23375e3b303/oss_fuzz_integration/get_full_coverage.py#L105-L116

if you open a PR with an initial suggestion then I'll be happy to assist

Navidem:
This is interesting @swirsz, thanks!

DavidKorczynski:
Fixed by https://github.com/ossf/fuzz-introspector/pull/318

PR #328, opened 2022-06-13
oss-fuzz: make multiple jobs optional

DavidKorczynski:
Ref: https://github.com/ossf/fuzz-introspector/issues/302


DavidKorczynski:
@swirsz we need to have the one core approach as default so I added a bit of logic around the multi-core running to make this possible. However, it's easy to use arbitrary amounts of jobs now, both by way of `get_full_coverage.py` and `run_both.sh. I also added a bit of documentation to it. This addresses the comment I put here: https://github.com/ossf/fuzz-introspector/pull/318#pullrequestreview-1002530897

PR #337, opened 2022-06-19
Update binutils dependency

fouzhe:
`texinfo` is the required dependency for command `makeinfo` when building binutils


DavidKorczynski (approved):
Thanks!

PR #339, opened 2022-06-19
Update local example script

fouzhe:
It seems this script is invoked in dir `work`:
https://github.com/ossf/fuzz-introspector/blob/22822c81cf2cfb0f6bd12e06ac8c14266a333ba4/README.md?plain=1#L189-L194
so it doesn't need `cd work` again.

### DIFF
--- a/tests/simple-example-0/build_cov.sh
+++ b/tests/simple-example-0/build_cov.sh
@@ -17,7 +17,9 @@
 #rm -rf ./work_cov
 #mkdir work_cov
 #cd work_cov
-cd work
+if [ -d "work" ]; then
+    cd work
+fi
 
 ../../../build/llvm-build/bin/clang -flegacy-pass-manager -fprofile-instr-generate -fcoverage-mapping -fsanitize=fuzzer -g ../fuzzer.c -o fuzzer
 ./fuzzer -max_total_time=3


DavidKorczynski:
I think the reason the `cd work` is there is so it works with https://github.com/ossf/fuzz-introspector/blob/main/tests/build_simple_example.sh

DavidKorczynski:
@fouzhe could you switch this change to be in the README instead of in `build_cov.sh`? Specifically, change it such that the following:
```
# Run compiler pass to generate *.data and *.data.yaml files
mkdir work
cd work

# Run script that will build fuzzer with coverage instrumentation and extract .profraw files
# and convert those to .covreport files with "llvm-cov show"
../build_cov.sh
```

Becomes

```
# Run compiler pass to generate *.data and *.data.yaml files
mkdir work

# Run script that will build fuzzer with coverage instrumentation and extract .profraw files
# and convert those to .covreport files with "llvm-cov show"
../build_cov.sh
cd work
```

I think that should make things work, right?

DavidKorczynski (changes requested):
Instead of deleting the line, could you wrap the `cd work` in an if-statement that checks if `work` is a directory, and if so navigates into it. I.e. the `cd work` command should only be used if there is a `work` directory. I think that should make things work with the instructions in README and also https://github.com/ossf/fuzz-introspector/blob/main/tests/build_simple_example.sh

grandnew:
Sounds reasonable and I've patched it.

DavidKorczynski (approved):
Thanks @fouzhe

PR #357, opened 2022-06-28
Report page: Improve UI

AdamKorcz:
1. Overall minor change of colors here and there.
2. Change styling of tables
3. Add button in left side with link to coverage

Preview:
fi-updates

### DIFF
--- a/post-processing/fuzz_html.py
+++ b/post-processing/fuzz_html.py
@@ -945,7 +945,7 @@ def create_html_report(
     ###########################
     # Fix up table of contents.
     ###########################
-    html_toc_string = fuzz_html_helpers.html_get_table_of_contents(toc_list)
+    html_toc_string = fuzz_html_helpers.html_get_table_of_contents(toc_list, coverage_url)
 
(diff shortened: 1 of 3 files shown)


DavidKorczynski:
looks sweet -- one thing we should look into adding here is a link to "per target coverage" as well. We have the per-target coverage in fuzz introspector data and I don't think this is readily accessible elsewhere: it would be great if we could get that.

Am not sure if it makes sense to have a similar button for per target coverage with some drop-down list with links to each fuzzer?

DavidKorczynski:
down the line I wonder if it makes sense to switch up to using Tabs -- for example, the introspector report is getting large and it may make sense to split it up by grouping related content together through tabs. Otherwise the report will grow very large as we progress. I think opening up for tabs may be nice because we can start to create very dedicated pages -- for example, I can see some value in having a dedicated page that simply tells a ton of data about a single function. In either case this is just some stream of thoughts and perhaps a bit irrelevant to this pr.

DavidKorczynski (approved):

PR #360, opened 2022-06-29
Use file path instead of file name in branch blocker class

### LINKED ISSUE #355: branch profiler: include full file path rather than filename?
@Navidem Taking a look at https://storage.googleapis.com/oss-fuzz-introspector/bind9/inspector-report/20220627/summary.json,

it looks like we have filenames specified like `"source_file_name": "lex.c"`, `"source_file_name": "buffer.c"`. For this to be automatable, we should store the full relative path from the project instead.

Navidem:
Fixes #355

### DIFF
--- a/post-processing/fuzz_analysis.py
+++ b/post-processing/fuzz_analysis.py
@@ -16,6 +16,7 @@
 
 import abc
 import logging
+import os
 
 from typing import (
     Dict,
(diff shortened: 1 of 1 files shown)


post-processing/fuzz_analysis.py
```diff
@@ -378,8 +377,9 @@ def detect_branch_level_blockers(
             continue
 
         llvm_branch_profile = functions_profile[function_name].branch_profiles
+        source_file_path = functions_profile[function_name].function_source_file
         # Just extract the file name and skip the path
-        source_file_name = functions_profile[function_name].function_source_file.split('/')[-1]
+        source_file_name = source_file_path.split('/')[-1]
```
oliverchang, line 382: Just use `os.path.basename` here?

Navidem: These are populated from debug info in an LLVM pass, so I think does not make sense to use os.path etc.

oliverchang: Could you elaborate a bit more? os.path.basename works generally any any path strings. 

It is however, tied to the OS it's running on. I.e. If this script runs on Windows on something generated from Linux, it might not work.

Navidem: My bad! I misunderstood your initial comment. Using `os.path.basename` now.

oliverchang (approved):

PR #361, opened 2022-06-30
Collect coverage hitcount diff for each blocked side

Navidem:
Collecting the coverage hitcount diff between the blocked side and not-blocked side of each branch. My idea is that this later can be used to decide which strategy to use for unblocking.

My intuition is that if the diff is low (like 10) then setting `focus-function` can be promising.  Otherwise, for example if the diff is thousands, then the fuzzer has already tried to hit the branch and we may not get much just by focusing on that function.


oliverchang (approved):
nice idea!

PR #367, opened 2022-06-30
Report page: add button with links to per-fuzzer coverage reports

AdamKorcz:
Adds a drop down button to select per-fuzzer coverage reports in the top left corner.

It might be a nice detail to add the coverage percentage after the fuzzer name if we have that available.

Previews:
btn1
btn2


DavidKorczynski (approved):

DavidKorczynski:
For reference, this was discussed here https://github.com/ossf/fuzz-introspector/pull/357

PR #372, opened 2022-07-05
Restructure code to make it more modular

DavidKorczynski:
This performs the first step of https://github.com/ossf/fuzz-introspector/issues/373 Specifically:
- moves the `llvm` folder into `frontends`. The point of this is down the line we will have multiple frontends (for other languages) and grouping them now improves structure.
- restructures `post-processing` into a folder `src` with a sub-folder `fuzz-introspector`, to help the Python hygiene
- extracts all datatypes (classes) out from `fuzz_data_loader.py` and into a separate folder `datatypes` split into individual files for each class e.g. a file for FuzzerProfile, FunctionProfile etc.
- removes `fuzz_` prefix from all Python files, as the new folder structure makes it clear what code belongs to fuzz introspector

Note that this does not have any implications in the workflow as such. There is a slight change in folder structure which needs to be applied in OSS-Fuzz (see the diff which does it).


DavidKorczynski:
@oliverchang @Navidem @AdamKorcz could you take a look at and let me know your thoughts on this.

DavidKorczynski:
oss-fuzz tests pass https://github.com/ossf/fuzz-introspector/blob/main/oss_fuzz_integration/test_projects.py

PR #380, opened 2022-07-07
tests: add unit testing set up

DavidKorczynski:
Adds set up for writing unit tests targeting the python code. Adds a test file for `utils.py`.

This is still early so am not sure if it really should be placed here or if we should reorganise the `/tests` folder and put it in there, or some third option. However this gives us a start.


AdamKorcz (approved):
LGTM

PR #401, opened 2022-07-19
Calltree page: Add overview

AdamKorcz:
Adds an overview of the calltree in the side. It reuses the image generated in the backend.

side-overview

Signed-off-by: AdamKorcz <Adam@adalogics.com>


DavidKorczynski:
Awesome -- next step is then syncing the position in the calltree with some identifier on top of the bitmap?

PR #405, opened 2022-07-20
Avoids redirection and removes extra /

Navidem:


DavidKorczynski (approved):
LGTM! Note that it will switch redirect and not avoid it: "tree" is used for directories and "blob" is used for files. E.g. the link here https://github.com/ossf/fuzz-introspector/blob/7981f883a1048e035ea942730799fc5b7c1a4199/src/fuzz_introspector/html_report.py#L528 links to https://github.com/ossf/fuzz-introspector/tree/main/doc/ and will not have any redirection to `blob`. the `tree` urls will be redirected to `blob` urls for files only

Issue #356, opened 2022-06-28
branch profiler: branches with 0 blocked_reachable_complexity and 0 blocked_not_covered_complexity?


oliverchang:
looking at https://storage.googleapis.com/oss-fuzz-introspector/arrow/inspector-report/20220627/summary.json, I see some branch blockers with:

`"blocked_not_covered_complexity": 0, "blocked_reachable_complexity": 0`

Is this correct? If they are, do they still make sense to be included in this list?

Navidem:
> Is this correct? If they are, do they still make sense to be included in this list?

This is not necessarily incorrect. What may happen is that the static analysis may fail to identify the callees or collect complexity for the identified callees. So the branch blocker has no source of information to compare such blocked branch sides, but it has identified the side is blocked as the coverage reports shows not hit for it. Therefore, I still want these cases to be reported as it is informative in terms of blockage.

That said, I will debug this specific case of `arrow` and post the update.

PR #409, opened 2022-07-22
Update total complexity of functions for each fuzzer

Navidem:
This is the culprit for #356.

What's happening is that the `total_cyclomatic_complexity` updates in `MergedProjectProfile` while branch blocker works with each `FuzzerProfile`. So we need the total complexity per fuzzer.


oliverchang:
Is this ready to be reviewed?

Navidem:
> Is this ready to be reviewed?

Having #411 merged, this one is not pursued, as explained in https://github.com/ossf/fuzz-introspector/issues/356#issuecomment-1194755658.

PR #421, opened 2022-07-30
Add branch blockers table to the portal

Navidem:


oliverchang:
Nice! Is there a screenshot?

Rather than adding a new table, would it make more sense to integrate this into the existing fuzz blocker tables instead?

Navidem:
> Nice! Is there a screenshot?

image


> Rather than adding a new table, would it make more sense to integrate this into the existing fuzz blocker tables instead?

This is the first attempt to show the branch blockers on portal mainly for @DavidKorczynski to test with it. Shortly we have to merge/replace blocker table.

oliverchang:
> Excited to see this in action! We can do such that there is only a single blocker table before bumping to OSS-Fuzz

Thanks, was making sure that's the end goal :)

In the meantime, I think there's value in rendering this table to help with evaluating the quality of these branch blockers on OSS-Fuzz. Could we hide this table by default, but have an "experimental" button we could click to show this?

Navidem:
> In the meantime, I think there's value in rendering this table to help with evaluating the quality of these branch blockers on OSS-Fuzz. Could we hide this table by default, but have an "experimental" button we could click to show this?

Check #422

Issue #312, opened 2022-06-09
Check Branch Blocker's report health


Navidem:

oliverchang:
Any details/examples of things that look broken?

DavidKorczynski:
> * for `wuffs` all fuzzers are getting same blockers reported

Am not sure if this is related, but it could perhaps be this: https://github.com/ossf/fuzz-introspector/issues/288

Also see: https://github.com/ossf/fuzz-introspector/issues/285

Navidem:
> Am not sure if this is related, but it could perhaps be this: #288
> 
> Also see: #285

I remembered these! And YES, those were smoking gun for at least `jsoncpp` case. The issue was the function names are used as dict key, and for cpp code we had to make sure we have names demangled all over coverage and llvm profile.

Navidem:
> for `wuffs` all fuzzers are getting same blockers reported

Now seeing the same for `jsoncpp`. The culprit is it fails to load per-fuzzer coverage. See #424

Navidem:
> > for `wuffs` all fuzzers are getting same blockers reported
> 
> Now seeing the same for `jsoncpp`. The culprit is it fails to load per-fuzzer coverage. See #424

This is fixed.

Navidem:
For at least two cases, I see a branch reported as blocked while the coverage report page show both side are hit:
- jsoncpp:
https://storage.googleapis.com/oss-fuzz-introspector/jsoncpp/inspector-report/20220802/fuzz_report.html#fuzz_blocker1
https://storage.googleapis.com/oss-fuzz-coverage/jsoncpp/reports-by-target/20220802/jsoncpp_proto_fuzzer/linux/src/jsoncpp/src/lib_json/json_value.cpp.html#L1409

Navidem:
> For at least two cases, I see a branch reported as blocked while the coverage report page shows both sides are hit:
> 
> * jsoncpp:
>   https://storage.googleapis.com/oss-fuzz-introspector/jsoncpp/inspector-report/20220802/fuzz_report.html#fuzz_blocker1
>   https://storage.googleapis.com/oss-fuzz-coverage/jsoncpp/reports-by-target/20220802/jsoncpp_proto_fuzzer/linux/src/jsoncpp/src/lib_json/json_value.cpp.html#L1409

#441 can help in fixing this.

Navidem:
[...]

PR #423, opened 2022-08-03
Fix function name inconsistency across coverage and profile

Navidem:
Fixed multiple issues with function name mangling causing branch blocker detection failure to load correct function profile. More details re https://github.com/ossf/fuzz-introspector/issues/312#issuecomment-1203274901

### DIFF
--- a/src/fuzz_introspector/analysis.py
+++ b/src/fuzz_introspector/analysis.py
@@ -415,7 +415,7 @@ def detect_branch_level_blockers(
         try:
             tmp_idx = branch_string.rfind(':')
             function_name = branch_string[:tmp_idx]
-            rest_string = branch_string[tmp_idx+1:]
+            rest_string = branch_string[tmp_idx + 1:]
             line_number, column_number = rest_string.split(',')
         except ValueError:
             logger.error(f"branch-profiling: error getting function name from {branch_string}")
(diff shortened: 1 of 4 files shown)


src/fuzz_introspector/datatypes/function_profile.py
```diff
@@ -68,3 +69,9 @@ def load_func_branch_profiles(
             bp_loaded[new_branch.branch_pos] = new_branch
 
         return bp_loaded
+
+    def load_func_names(self, input_list: List[str]) -> List[str]:
```
oliverchang, line 73: Rather than copy and pasting, would it make sense to put this into a helper module (e.g. utils) instead?

src/fuzz_introspector/analysis.py
```diff
@@ -413,14 +413,16 @@ def detect_branch_level_blockers(
 
         # Catch exceptions in case some of the string splitting fails
         try:
-            function_name, rest_string = branch_string.split(':')
+            tmp_idx = branch_string.rfind(':')
```
oliverchang, line 416: Can we just use rsplit instead, with maxsplit=1 ? 

https://docs.python.org/3/library/stdtypes.html#str.rsplit

Navidem: Done.

src/fuzz_introspector/datatypes/function_profile.py
Navidem: good call.

oliverchang (approved):

PR #445, opened 2022-08-08
Check for the line hit for fall through branch sides

### LINKED ISSUE #441: Provide a way to check line hit when coverage type is not file
The function is its current implementation does not do what it is supposed to for C/C++ coverage.

More specifically `self._cov_type` is not set and `self.file_map` is not populated for such coverages to be used here.
I can take a pass on it on Monday.

Navidem:
In a C/C++ coverage report the line info is categorized based on the function name. This PR checks if a line in a given function is hit or not.

Fixes #441

### DIFF
--- a/src/fuzz_introspector/code_coverage.py
+++ b/src/fuzz_introspector/code_coverage.py
@@ -185,6 +185,20 @@ class CoverageProfile:
         lines_hit = [ht for ln, ht in self.covmap[fuzz_key] if ht > 0]
         return len(self.covmap[fuzz_key]), len(lines_hit)
 
+    def is_func_lineno_hit(self, func_name: str, lineno: int) -> bool:
+        """
+        Checks if a given line number in a function is hit.
+        """
+        func_hit_details = self.get_hit_details(func_name)
+
+        for line_info in func_hit_details:
+            if lineno == line_info[0]:
+                if line_info[1] != 0:
+                    return True
+                else:
+                    return False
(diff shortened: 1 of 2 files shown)


src/fuzz_introspector/code_coverage.py
```diff
@@ -221,6 +235,7 @@ def load_llvm_coverage(
                 found_name = target_name
 
     cp = CoverageProfile()
+    cp.set_type("file")
```
DavidKorczynski, line 238: This should be "function" or something of this nature, but not "file", right?

Navidem: Oops my bad, this was intended to be removed. Thanks for pointing out.

Btw, `function` type sounds good. I will use it.

DavidKorczynski: awesome!

DavidKorczynski (approved):
lgtm!

Down the line I think it does make sense to make the code coverage API cleaner as described here https://github.com/ossf/fuzz-introspector/issues/441#issuecomment-1207196169 to avoid too many `if` statements across the code. This could be enforced if we require that the coverage type is specified for each `CoverageProfile`.

Navidem:
> This could be enforced if we require that the coverage type is specified for each `CoverageProfile`.

+1 on enforcing coverage type in `CoverageProfile`. This will help as we move forwards with different langs.

PR #451, opened 2022-08-11
oss-fuzz: exclude llvm-symbolizer

evverx:
to prevent the script from running it as a fuzz target and getting
```
Running llvm-symbolizer
...
/out/llvm-symbolizer -rss_limit_mb=2560 -timeout=25 -max_total_time=20 -detect_leaks=0 /tmp/llvm-symbolizer_corpus < /dev/null^M
error: unknown argument '-r'^M
error: unknown argument '-_'^M
error: unknown argument '-l'^M
error: unknown argument '-m'^M
...
Executing finished with exception
```

### DIFF
--- a/oss_fuzz_integration/get_full_coverage.py
+++ b/oss_fuzz_integration/get_full_coverage.py
@@ -56,11 +56,15 @@ def get_fuzzers(project_name):
     execs = []
     for l in os.listdir("build/out/%s"%(project_name)):
         print("Checking %s"%(l))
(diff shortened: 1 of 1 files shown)


DavidKorczynski (commented):
Ah nice, thanks! While at it would you mind adding these as well? https://github.com/google/oss-fuzz/blob/800a54e987ca7f0eb86f733d220e4eb7f5d80d3d/infra/base-images/base-runner/coverage#L24-L28

evverx:
Not at all. Done.

DavidKorczynski (approved):
Awesome, thanks!

PR #453, opened 2022-08-11
oss-fuzz: no longer launch the server showing coverage reports

evverx:
to make it easier to run the script without having to press Ctrl-C
every time.

https://github.com/google/oss-fuzz/commit/218f5dc3cf70bd6


Navidem (approved):
Thanks!

Issue #24, opened 2022-01-07
Please provide a real world example


vanhauser-thc:
The README does not tell how this is to be used with a project and instead points to a build_all script, which then needs to be further analyzed how it is used. This is not a good user experience.

A simple "this is how you need to build your target" and "this is how you use the Python script to build the report" in the README would be a much lower cost for users. 

:)

DavidKorczynski:
Thanks, much appreciated! I will get something like this up

Meanwhile, the recommended way to test this is using the OSS-Fuzz approach (https://github.com/ossf/fuzz-introspector#testing-with-oss-fuzz). It is the recommended way because it's through there we currently support combining runtime coverage data with the compiler plugin. The drawback is that it will override your existing OSS-Fuzz Docker images.

vanhauser-thc:
as you can imaging I will not use this with oss-fuzz :) (and several other people either) hence the issue ...

DavidKorczynski:
Yeah makes sense, I have this in mind for the near future but will just add one or two things in the compiler plugin to make the UX a bit nicer. I will do this in the coming days!

PR #454, opened 2022-08-12
oss-fuzz: cleanup diff

DavidKorczynski:
Remove commenting out starting the coverage web server as it's actually not needed: Ref: https://github.com/ossf/fuzz-introspector/pull/453

### DIFF
--- a/oss_fuzz_integration/oss-fuzz-patches.diff
+++ b/oss_fuzz_integration/oss-fuzz-patches.diff
@@ -159,13 +159,6 @@ index 1659ee94..fc5e682d 100755
  
    # Create an empty summary file for now
    echo "{}" >> $SUMMARY_FILE
-@@ -392,5 +399,5 @@ if [[ -n $HTTP_PORT ]]; then
-   # Serve the report locally.
-   echo "Serving the report on http://127.0.0.1:$HTTP_PORT/linux/index.html"
-   cd $REPORT_ROOT_DIR
--  python3 -m http.server $HTTP_PORT
-+  #python3 -m http.server $HTTP_PORT
- fi
 diff --git a/projects/pyyaml/build.sh b/projects/pyyaml/build.sh
 index 98f406ad..62485ae4 100644
 --- a/projects/pyyaml/build.sh


AdamKorcz (approved):

evverx:
I wonder how this diff gets updated? For example I got rid of confusing messages like
```
cp: '/src/inspector/fuzzerLogFile-0-e9rZnlKm9n.data.yaml' and '/src/inspector/fuzzerLogFile-0-e9rZnlKm9n.data.yaml' are the same file
```
by editing `infra/base-images/base-builder/Dockerfile` but I'm not sure where that patch should go. I suspect it should be included in this diff but I'm not sure how it all works :-)

DavidKorczynski:
The flow of the diff is something along the lines of supporting everything we need while developing on Fuzz Introspector until it's ready to get merged with OSS-Fuzz, and when that happens we clean up the diff based on what was added in the OSS-Fuzz bump.

Feel free to push things to the diff, e.g. if it eases workflow or it could be useful to use in an OSS-Fuzz environment -- we're happy with accumulating things in there that are useful!

evverx:
Got it. Thanks!

> when that happens we clean up the diff based on what was added in the OSS-Fuzz bump

As far as I understand it kind of makes it impossible to follow the main branch of the FI repository because changes to OSS-Fuzz can break FI until the diff is updated. At least I can't seem to figure out how https://github.com/ossf/fuzz-introspector/commit/b7ee178347f65ed434e3418c4818f09c388e82a7 can be backported to the base-builder:introspector image without temporarily breaking scripts like `build_post_processing.sh` that expect that patch to be applied cleanly to the OSS-Fuzz repository. More generally I think it's kind of hard to keep it all in sync without running into issues like https://github.com/ossf/fuzz-introspector/issues/476 (in retrospect I should have rebuilt the images locally when https://github.com/ossf/fuzz-introspector/pull/461 landed to get the python code to work).

DavidKorczynski:
That's good to know @evverx I feel in general this improved user experience also ties back to https://github.com/ossf/fuzz-introspector/issues/24

In short, we're pushing towards making this more accessible and reducing the efforts to get started with FI and contributing to it. The issues you posted recently have been a great help identifying relevant areas to resolve.

evverx:
I think the problem is that there are roughly speaking three different use cases and they kind of conflict with each other. OSS-Fuzz users I assume would be interested in building projects using the OSS-Fuzz toolchain without having to clone the FI repository (or maybe even know how it works under the hood). I think ideally something like
```sh
./infra/helper.py build_fuzzers --sanitizer introspector project
./infra/helper.py check_build --sanitizer introspector project
./infra/helper.py introspector project
```
should be enough to build FI reports there. People developing FI can't get by using the OSS-Fuzz repository only so they should somehow keep it all in sync. I think https://github.com/ossf/fuzz-introspector/issues/24 should cover the third use case where projects that aren't integrated into OSS-Fuzz try to experiment with FI. I have to admit I'm not sure how to cover all the use cases.

DavidKorczynski:
Yeah, the different set of use cases does make maintenance and user experience a bit difficult and time-consuming if fully pursued. We've had some requests for non OSS-Fuzz use cases and the plan is to support this. Until now, we've had almost all attention on OSS-Fuzz and supporting it in a manner where the user is not responsible for running/handling FI itself. This is changing (and has for some time), and we intend on better supporting all of the above use cases. However, I think it's reasonable to assume OSS-Fuzz support will remain top priority for some time.

evverx:
> we've had almost all attention on OSS-Fuzz and supporting it in a manner where the user is not responsible for running/handling FI itself

I think it makes sense because FI reports built automatically on OSS-Fuzz make it easier to start using FI (assuming projects can be built with it automatically). If they can't maintainers should somehow reproduce and ideally fix failures locally and the process isn't as smooth as it could probably be: https://github.com/ossf/fuzz-introspector/issues/483. More generally as I see it the idea behind FI is to help developers to improve their fuzz targets and the OSS-Fuzz workflow where they need to push their changes and wait for OSS to build the reports isn't suitable for that. The again, I agree that it's nice to have those reports anyway.

> This is changing (and has for some time), and we intend on better supporting all of the above use cases.

Good to know. Thanks!

> However, I think it's reasonable to assume OSS-Fuzz support will remain top priority for some time.

Fair enough.

Issue #447, opened 2022-08-09
analysis plugin: add API-finder plugin


DavidKorczynski:
It would be nice to have an API finder plugin that can search for occurrences of given library functions in the code. This could be matched with runtime information to assess all occurrences are hit or not.

--------------------------------------
As an example, consider the following fuzzer + target:
```c
#include <string.h>
#include <stdio.h>
#include <stdlib.h>
#include <stdint.h>

void execute_sys1(char *directory) {
  if (strlen(directory) > 70) {
    return;
  }
  char cmd[80];
  sprintf(cmd, "ls -la %s", directory);
  system(cmd);
}

void execute_sys2(char *directory) {
  if (strcmp(directory, "/tmp/") == 0) {
    char cmd[80];
    sprintf(cmd, "ls -la %s", directory);
    system(cmd);
  }
}

void execute_sys3(char *cmd) {
  system("ls -la /tmp/");
}

void execute_sys4(char *directory) {
  if (strlen(directory) > 70) {
    return;
  }
  char cmd[80];
  sprintf(cmd, "%s", directory);
  system(cmd);
}

void check_directory(char *directory, int api) {
  if (api == 1) {
    execute_sys1(directory);
  }
  else if (api == 2) {
    execute_sys2(directory);
  }
  else if (api == 3) {
    execute_sys3(directory);
  }
  // Intentionally commented out
  //else {
  //  execute_sys4(directory);
  //}
}

int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
  if (size < 5) {
	  return 0;
  }
  char *s1 = (char*)malloc(size+1);
  memcpy(s1, data, size);
  s1[size] = '\0';

  check_directory(s1, 2);
  check_directory(s1, 3);
  return 0;
}
```

Assuming the function of interest is `system`, then the analysis should be able to tell:
1) There are four occurrences of `system`
2) The occurrences of `system` in `execute_sys1`, `execute_sys2`, `execute_sys3` are all statically reachable
3) The occurrences of `system` in `execute_sys2` and `execute_sys3` are covered by the fuzzer at runtime
4) The occurrence of `system` in `execute_sys1` is not covered by the fuzzer at runtime.
5) The occurrence of `system` in `execute_sys4` is not statically reachable and not covered by the fuzzer at runtime.

A table showing all occurrences of a given API and then columns with source code location + if it's statically reachable + if it's covered at runtime would be the best for now I recon.

[...]

PR #456, opened 2022-08-12
New C testing project for fuzz-introspector with oss-fuzz

arthurscchan:


DavidKorczynski (changes requested):
Could you rename folder `tmp-proj` to `oss-fuzz-test-projects` and also add header license to all files?

DavidKorczynski (changes requested):
We have a different licence (copyright notice specifically) from OSS-Fuzz.

For scripts:  https://github.com/ossf/fuzz-introspector/blob/6d1e0a5a67370c9d26da86a484cb5b8311d9480f/oss_fuzz_integration/build_post_processing.sh#L1-L16

 and for C code:
 
 https://github.com/ossf/fuzz-introspector/blob/6d1e0a5a67370c9d26da86a484cb5b8311d9480f/frontends/llvm/lib/Transforms/FuzzIntrospector/FuzzIntrospector.cpp#L1-L14
 
 with different years though

DavidKorczynski (changes requested):
Could we place the project in a specific folder inside of oss-fuzz-test-projects? In the anticipation we will have more, could you locate it in a folder called "injection-test" inside of the "oss-fuzz-test-projects" folder?

DavidKorczynski:
Ref: https://github.com/ossf/fuzz-introspector/issues/447

DavidKorczynski (approved):

Issue #452, opened 2022-08-11
enable download public OSS-Fuzz corpus when running locally


DavidKorczynski:
When running locally using `run_both.sh` and `get_full_coverage.py` it relies on collecting runtime coverage. To do this the current approach is running the fuzzers for X number of seconds.

A more complete approach is to run it with the OSS-Fuzz public/private corpus. This will also make things easier for debugging when improving fuzzers by way of Fuzz Introspector.

tldr: enable use of pre-existing corpus and also downloading of most recent OSS-Fuzz public corpus.

Ref: https://github.com/google/oss-fuzz/pull/8222#issuecomment-1212553623
Ref: https://github.com/ossf/fuzz-introspector/issues/302#issuecomment-1147981215

evverx:
In the meantime it should be possible to bring for example public OSS-Fuzz corpora by editing build scripts and using those corpora as seed corpora. As far as I can remember I also tried to put them in `build/corpus/<project>/<fuzz_target>/` (where `helper.py coverage` expects to find them) but it didn't work out because as far as I can see they are wiped out by `get_full_coverage.py` https://github.com/ossf/fuzz-introspector/blob/bdf2907db1e80d98a98c6eccf39f88ae1289986c/oss_fuzz_integration/get_full_coverage.py#L153-L160

evverx:
> This will also make things easier for debugging when improving fuzzers by way of Fuzz Introspector.

I think for that to fully work apart from corpora it should be possible to point the script to directories where fuzz targets are (by analogy with `./infra/helper.py build_fuzzers project /path/to/source/code`). So far I have been playing with projects keeping their fuzz targets in the oss-fuzz repository and there the oss-fuzz repository is enough but if I switched to projects that integrated their fuzz targets into their repositories I don't think there would be an easy way to debug them.

FWIW it appears "per-fuzzer coverage" is hidden behind "table of contents" with more than one fuzz target
![Screenshot 2022-08-13 at 04 59 03](https://user-images.githubusercontent.com/10195800/184464322-c5f6de32-9459-4b56-b7e8-c2560190e95b.png)

PR #459, opened 2022-08-13
oss-fuzz: allow setting ports explicitly

evverx:
to avoid collisions in general and make it easier to run the script in VMs/containers where only certain ports are forwarded.

I went with the environment variable because I wasn't sure how to keep the script working with `project time --jobs=...`.

I also dropped a redundant command. It isn't run anyway because the server is either interrupted with Ctrl-C
or fails with something like "OSError: [Errno 98] Address already in
use".


DavidKorczynski (approved):
lgtm! re env variable, if we extend this further down the line it may be the best to rely on flags rather than environment vairables, but let's tackle that when time comes. I assume https://github.com/ossf/fuzz-introspector/issues/452 might cause some more flags to be needed.

PR #460, opened 2022-08-14
build_all: make the custom clang buildable again

evverx:
and also update doc/LocalBuild.md.

Some scripts are still out of sync with the new layout and the documentation but they can probably be fixed later.

### DIFF
--- a/build_all.sh
+++ b/build_all.sh
@@ -14,6 +14,8 @@
 #
 ################################################################################
 
+set -ex
+
 BASE=$PWD
 mkdir build
 cd build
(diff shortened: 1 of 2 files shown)


DavidKorczynski (approved):
Thanks! Am working on this as writing so will push some more updates shortly (some updates are needed in the `tests/` folder).

evverx:
FWIW I think in terms of supporting for example incremental builds or working with two snapshots of the same project it would be great if there was a way to pull images with the custom clang and FI directly instead of building it manually or pulling the whole OSS-Fuzz toolchain. It's kind of possible to use `gcr.io/oss-fuzz-base/base-builder:introspector` where a bunch of python dependencies have to be installed manually (and `FUZZINTRO_OUTDIR` has to be unset) but I wouldn't say it's straightforward.

PR #469, opened 2022-08-15
main.py: a few fixes related to logging

evverx:
* set up the logger before printing anything
* no longer print the same message twice
* avoid printing "Running ..." when the script is run with just `--help`


DavidKorczynski (approved):
Good stuff, thanks!

PR #473, opened 2022-08-16
plugins: refine function call analyser

DavidKorczynski:
Adds filtering in function call analyser to look for specific sinks.


AdamKorcz (approved):

DavidKorczynski:
Screenshot from 2022-08-16 21-54-16

 Example of using it to find the `wordexp` call from https://github.com/syoyo/tinygltf/issues/368

PR #474, opened 2022-08-16
Fix bug in function profile changes

arthurscchan:
Fix bug for creating new Third Party API call coverage analyser.
Solve a bug for: #447


AdamKorcz (approved):
LGTM

Issue #476, opened 2022-08-16
FI seems to be failing with "KeyError: 'Callsites'"


evverx:
```sh
../run_both.sh dbus-broker 60
...
Traceback (most recent call last):
  File "/fuzz-introspector/src/main.py", line 141, in <module>
    main()
  File "/fuzz-introspector/src/main.py", line 122, in main
    return_code = commands.run_analysis_on_dir(
  File "/fuzz-introspector/src/fuzz_introspector/commands.py", line 53, in run_analysis_on_dir
    profiles = data_loader.load_all_profiles(target_folder, language)
  File "/fuzz-introspector/src/fuzz_introspector/data_loader.py", line 152, in load_all_profiles
    profile = read_fuzzer_data_file_to_profile(data_file, language)
  File "/fuzz-introspector/src/fuzz_introspector/data_loader.py", line 61, in read_fuzzer_data_file_to_profile
    FP = fuzzer_profile.FuzzerProfile(cfg_file, data_dict_yaml, language)
  File "/fuzz-introspector/src/fuzz_introspector/datatypes/fuzzer_profile.py", line 69, in __init__
    self._set_function_list(frontend_yaml)
  File "/fuzz-introspector/src/fuzz_introspector/datatypes/fuzzer_profile.py", line 405, in _set_function_list
    func_profile = function_profile.FunctionProfile(elem)
  File "/fuzz-introspector/src/fuzz_introspector/datatypes/function_profile.py", line 55, in __init__
    self.callsite = self.load_func_callsites(elem['Callsites'])
KeyError: 'Callsites'
The following command is about to be run to start a webserver: cd ./corpus-0/inspector-report/ && python3 -m http.server 8008
```
What's weird is that even though FI failed its exit code was 0 apparently and the script proceeded to launch a web server.

DavidKorczynski:
Thanks! This is a nasty regression, I'll fix it up shortly.

PR #477, opened 2022-08-16
function profile: quick fix

DavidKorczynski:
Fixes: https://github.com/ossf/fuzz-introspector/issues/476

The issue happens when the underlying clang is out of sync with the post
processing part, because the "Callsites" element is not available in the
older clang frontend. We should, however, support a more safe loading
anyways.

### DIFF
--- a/src/fuzz_introspector/datatypes/function_profile.py
+++ b/src/fuzz_introspector/datatypes/function_profile.py
@@ -52,7 +52,10 @@ class FunctionProfile:
         self.branch_profiles = self.load_func_branch_profiles(elem['BranchProfiles'])
 
         # Saving callsites for this function
-        self.callsite = self.load_func_callsites(elem['Callsites'])
+        try:
+            self.callsite = self.load_func_callsites(elem['Callsites'])
+        except Exception:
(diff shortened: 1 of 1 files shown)


AdamKorcz (approved):

evverx:
I'm also not sure why the exit code was 0 there but it's another issue.

DavidKorczynski:
> @DavidKorczynski thanks! FI no longer crashes as far as I can see. I wonder if it should crash (or at least log something) if the "ThirdPartyAPICoverageAnalyser" analysis is on? As far as I understand without callsites the analysis doesn't make much sense.

Ah cool!

The specific callsite you met was slightly different to what has been in FI for a while. Prior, we only collected callsites for code that was deemed reachable from the fuzzer. We also had a slight notion of this before in that we had data on which functions a given function would reach -- although we didn't save the callsites if they were not in the CFG of the fuzzer explicitly.

The new stuff is all callsites across the whole program.

DavidKorczynski:
> I'm also not sure why the exit code was 0 there but it's another issue.

Saw that one too! I agree it should not be the case

Issue #468, opened 2022-08-15
There doesn't seem to be a way to embed ids into FI reports


evverx:
I'm not sure how exactly FI is supposed to be used but it seems to me that in general building at least two reports and comparing them would be a reliable way to figure out whether anything has been improved. To make it easier to tell reports apart it would great if there was a way to embed SHA1s and/or dates into reports. Dates and SHA1s would be useful on OSS-Fuzz too to make it clear that for example https://storage.googleapis.com/oss-fuzz-introspector/systemd/inspector-report/20220422/fuzz_report.html is out of date and shouldn't be relied on (The date is included in the link there but it took me a while to find it and I think it would be better if it was in the report itself).

More generally I think it would be ideal if FI could generate diffs of some kind to make it easier to compare two reports by analogy with how for example https://coveralls.io/github/systemd/systemd?branch=main works but I think for the time being SHA1s/dates should at least make it easier to figure out where reports come from.

MarahGamdou:
Hi, 
I'm wondering if the SHAs have been included in the reports or if it is possible to include them. If not, do you think the date of report generation can be reliable to deduce SHAs? 
Thank you!

DavidKorczynski:
We have dates in the report now, which can be used for tracking.

We can add sha's of commits to e.g. the metadata section. Not all projects rely on git or version control systems as such (xpdf if I remember correctly) so I think it's smartest to keep it as an optional.

Diffing two reports is useful -- perhaps a way to start this is to include more summary and easy-to-digest data at the top of the report.

MarahGamdou:
Thanks! 
I agree that this may not be useful for all projects, so keeping this optional is better. I also think sha's of commits can be very useful to track the exact code that was used to generate the reports. Sha's of commits are also integrated in oss-fuzz reports now (in the fuzzer logs and bug reports). Do you think it's somehow possible to get it from oss-fuzz reports and integrate it in the FI metadata for better tracking? If not, what would you suggest? 
Thank you for your help :))

evverx:
@MarahGamdou I'm not sure if it helps but as a stopgap (assuming the projects you're interested in use `git`) it should be possible to embed SHAs into FI build logs by adding something like
```sh
printf 'SHA: %s\n' $(git rev-parse HEAD)
```
[...]

PR #487, opened 2022-08-24
html_report: add report generation timestamp

DavidKorczynski:
Ref: https://github.com/ossf/fuzz-introspector/issues/468

### DIFF
--- a/src/fuzz_introspector/html_report.py
+++ b/src/fuzz_introspector/html_report.py
@@ -21,6 +21,7 @@ import typing
 import bs4
 import random
 import string
+from datetime import datetime
 import matplotlib.pyplot as plt
 
 from matplotlib.patches import Rectangle
@@ -840,6 +841,7 @@ def create_html_report(
     # Start creation of core html
     html_body_start = '<div class="content-section">'
     html_overview = "<div class=\"report-box\">"
+    html_overview += "<b>Report generation date:</b>" + datetime.today().strftime('%Y-%m-%d %H:%M:%S') + "<pr>"
     html_overview += html_helpers.html_add_header_with_link(
         f"Project overview: {report_name}",
         1,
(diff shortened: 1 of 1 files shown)


Navidem (approved):

DavidKorczynski:
Screenshot from 2022-08-24 19-06-21

Navidem:
> Screenshot from 2022-08-24 19-06-21

Nice!

evverx:
Thanks! To judge from
Screenshot 2022-08-25 at 14 40 40
it seems to show dates only. I tend to generate a few reports and view them at the same time so I'd probably add `%H:%M:%S` there to make it easier to tell them apart. I can send a PR if it makes sense.

DavidKorczynski:
> I tend to generate a few reports and view them at the same time so I'd probably add `%H:%M:%S` there to make it easier to tell them apart. I can send a PR if it makes sense.

If you find it useful to have the precise timestamp then I'd be happy to have it and accept the PR!

I was a bit on the fence about this the main thing was once I reviewed the code I thought we should add a timezone if displaying the precise timestamp. The timezone required an extract package and I wasn't entirely sure what the hours+sec would add from a user perspective as for local runs I use the counter on the `corpus-` dirs to identify "when it was run". That said, I don't have any strong opinion on this and am happy to accept precise timestamp without timezone too!

evverx:
> I reviewed the code I thought we should add a timezone if displaying the precise timestamp

That's a good point. I'd just probably go with UTC but as far as I can tell it wouldn't be consistent with other dates on OSS-Fuzz. Let's put this idea on hold :-)

> I use the counter on the corpus- dirs to identify "when it was run"

I run FI in containers and forward ports to a VM where they are forwarded to a real machine :-) so in my case it's easier to memorize ports.

Issue #485, opened 2022-08-22
FI reports should probably include direct links to *.covreport, *.data, *.data.yaml and build logs


evverx:
Trying to figure out why FI showed [a warning saying that "The number of covered functions are larger than the number of reachable functions"](https://storage.googleapis.com/oss-fuzz-introspector/elfutils/inspector-report/20220817/fuzz_report.html#functions_cov_hit_1) I ended up building `elfutils` with FI locally, parsing the "covreport" and "data" files and comparing the results. Once I got the list of the functions missing from the "data" file it was kind of obvious that it was due to https://github.com/ossf/fuzz-introspector/blob/main/doc/Calltree.md#indirect-pointers-makes-approximate-calltrees. I think ideally it would be great if FI could get diffs like that itself to make it easier to make sense of warnings like that. More generally I think it should be easy to get all the data used to produce FI reports.

As far as I understand It's kind of possible to download those files with
```diff
wget https://storage.googleapis.com/oss-fuzz-introspector/elfutils/inspector-report/20220817/fuzz-libelf.covreport
wget https://storage.googleapis.com/oss-fuzz-introspector/elfutils/inspector-report/20220817/fuzzerLogFile-0-Y1o6gbyVfs.data
```
but those links have to be inferred manually at this point.

Issue #176, opened 2022-04-02
include config and logging in final report


DavidKorczynski:
include configs and logs in the final report

This is useful to track, for instance, which functions were discarded from the analysis

DavidKorczynski:
Somewhat related: https://github.com/ossf/fuzz-introspector/issues/468 We need improved metadata in the UI which can be used to trace/track/compare reports.

evverx:
> We need improved metadata in the UI which can be used to trace/track/compare reports

Agreed. I think configs should be included in reports because they should be more or less the same everywhere but links to logs would depend on where reports were built. Usually it's OSS-Fuzz but FI can be used outside of OSS-Fuzz too and I think it would be better if optional links like that could be configurable by passing them explicitly somewhere or something like that.

evverx:
FWIW I think that even though FI reports are the main way to interact with FI it could (in theory) be used to automate issues like https://github.com/google/oss-fuzz/issues/7190 (where two snapshots of the same project should be compared). That particular issue could have been found by comparing coverage collected at runtime but combined with the static analysis it would be much more useful in general because it would cover scenarios where coverage is seemingly the same (or maybe even a bit better) but it can't keep up with new code that is suddenly reachable but isn't covered at all. I'm just spitballing but all in all I'm not sure FI should limit itself to reports on OSS-Fuzz (though I understand that it's the main use case) :-)

PR #490, opened 2022-08-25
analysis plugins: add metadata plugin

DavidKorczynski:
Display the data used for constructing the fuzzing report.

Includes links to .covreport, .data, .data.yaml

Ref: https://github.com/ossf/fuzz-introspector/issues/485

Build logs will come at a later stage.

### DIFF
--- a/src/fuzz_introspector/analyses/metadata.py
+++ b/src/fuzz_introspector/analyses/metadata.py
@@ -1,0 +1,105 @@
+# Copyright 2022 Fuzz Introspector Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


DavidKorczynski:
Screenshot from 2022-08-25 17-05-39

AdamKorcz (approved):

evverx:
It seems to be working. Thanks! Given that build logs are tracked in https://github.com/ossf/fuzz-introspector/issues/176 I think it's safe to say that https://github.com/ossf/fuzz-introspector/issues/485 is fixed. I'll go ahead and close it. I still think that it would be great if FI could show covered functions that aren't included in call graphs but it's a separate issue.

DavidKorczynski:
> I still think that it would be great if FI could show covered functions that aren't included in call graphs but it's a separate issue.

I think that would be real nice. Will fix this up

evverx:
> The goal is to use the runtime information to extend the callgraph whenever we observe something like this, because it's a fact whatever is executed is in essence within reach of the fuzzer

I agree they should probably be combined one way or another but I think in some cases it would mask build issues that should probably be fixed. For example some parts of call trees get lost because fuzz targets are linked against dependencies that have been compiled differently or they can be linked against shared libraries. Coverage data can be used to somewhat restore call trees there but fuzz targets like that still need fixing.

PR #495, opened 2022-08-30
oss-fuzz: make it possible to specify corpus

DavidKorczynski:
Adds the possibility to specify --corpus-dir to get_full_coverage.py.

Example usage: `../run_both.sh htslib 5 --corpus-dir=./corpus-11`

The folderp pointed to by --corpus-dir must have the same structure as
it would under build/corpus/proj_name/..

PR #496, opened 2022-08-30
oss-fuzz: substitute run_both.sh and get_full_coverage for a new manager

DavidKorczynski:
Follow up to: https://github.com/ossf/fuzz-introspector/pull/495

Specifically, first step towards a more general set up as described here https://github.com/ossf/fuzz-introspector/pull/495#issuecomment-1231547026


AdamKorcz (approved):

Navidem (commented):
This is nice! Thanks!

oss_fuzz_integration/README.md
```diff
@@ -29,7 +29,7 @@ From within this directory, run the commands:
 
 # Test a project
 cd oss-fuzz
-../run_both.sh htslib 20
+python3 ../manager.py full htslib 20
```
Navidem, line 32: nit: should we call this `runner.py`? and the options be like `runner.py introspector` vs. `runner.py coverage`?

I think this will be more intuitive, as this script is about running OSS-fuzz integration rather managing things. WDYT?

DavidKorczynski: > nit: should we call this `runner.py`? and the options be like `runner.py introspector` vs. `runner.py coverage`?

I'm happy to either way! one issue to overcome in that context is to avoid confusion in that `introspector` e.g. does both (1) build with asan; (2) run using asan fuzzers; (3) build coverage; (4) run coverage and finally (5) run introspector -- as such calling it `introspector` may be confusing? I called it `full` atm for that reason, although admittedly it's not super descriptive.

Navidem: > to avoid confusion in that `introspector` e.g. does

I wouldn't worry much about it honestly. I think the user should know the implicit meaning of running `introspector` includes running fuzzer and coverage. Such implicitly is already in `coverage` run, i.e we don't worry to mention that running for coverage needs the fuzzer be run before and corpus be available.

PR #501, opened 2022-09-02
Fix location of main.py for simple test

JordyZomer:
For this test, the location of main.py is in src instead of post-processing (post-processing doesn't exist in the main branch).

Signed-off-by: Jordy Zomer <jordy@pwning.systems>


DavidKorczynski (approved):
Ah, thanks a lot!

Navidem:
Thanks for the fix @JordyZomer!

PR #505, opened 2022-09-08
Merge branch blockers table

Navidem:
The logic checks for the presence of branch blocker, if the project does not support such capability it falls back to display calltree-based blockers.

### DIFF
--- a/src/fuzz_introspector/analyses/calltree_analysis.py
+++ b/src/fuzz_introspector/analyses/calltree_analysis.py
@@ -192,7 +192,7 @@ class Analysis(analysis.AnalysisInterface):
         the line it makes sense to have an easy wrapper for other HTML pages too.
         """
         complete_html_string = ""
-
+        blocker_infos = {}
         # HTML start
         html_header = html_helpers.html_get_header(
             calltree=True,
@@ -205,7 +205,11 @@ class Analysis(analysis.AnalysisInterface):
         if profile.branch_blockers:
             blockers_node_map = self.collect_calltree_nodes(profile.branch_blockers[:12],
                                                             profile.function_call_depths)
-            fuzz_blocker_nodes = list(blockers_node_map.values())
+            # Record the link to coverage report for the branch blocker.
+            for b_blocker, ct_node in blockers_node_map.items():
+                idx = self.create_str_node_ctx_idx(str(ct_node.cov_ct_idx))
+                blocker_infos[idx] = b_blocker.coverage_report_link
+
             fuzz_blocker_table = self.create_branch_blocker_table(
                 profile,
                 [],
@@ -224,6 +228,11 @@ class Analysis(analysis.AnalysisInterface):
                 "",
                 fuzz_blockers=fuzz_blocker_nodes
             )
+
+            for node in fuzz_blocker_nodes:
+                # The link to coverage report is not present in this type of blockers.
+                blocker_infos[self.create_str_node_ctx_idx(str(node.cov_ct_idx))] = ""
+
         if fuzz_blocker_table is not None:
             complete_html_string += "<div class=\"report-box\">"
             complete_html_string += "<h1>Fuzz blockers</h1>"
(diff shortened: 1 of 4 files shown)


Navidem:
A sample report can be found [here](https://storage.googleapis.com/oss-fuzz-introspector-testing/htslib/inspector/fuzz_report.html#fuzz_blocker0).

oliverchang:
(Can be a followup PR). Would it make sense to add this link to the call tree view itself? I.e. have a link to the blocking branch in addition to call site / function?

DavidKorczynski:
The precision of the blockers with respect to the calltree is not ideal. For example, the 5 blockers in the bottom of the blocker table points to the same location in the calltree (calltree index 1929) but with slightly different offsets in the source code:
Screenshot from 2022-09-09 10-33-27

The blocker itself is also a bit counter-intuitive when inspecting it:

Screenshot from 2022-09-09 10-34-19

I think this happens because the branch blockers are each affiliated with the source code location of various conditions in the `hts_open_format` but the pairing between branch blocker and calltree is with function name rather than more precise location. It would be nice to refine this -- perhaps either perhaps mixing in with the existing blockers if some hybrid can be done or matching based on more details of the node e.g. source location.

In the above context, could we do something to avoid reporting the multiple branch blockers in `hts_open_format` that are quite similar in nature? For example, I suspect the branch blockers in `hts_open_format` that each have 7330 or 7326 reachable complexity most likely have a large intersection in terms of the code they block.


Looking more at the blockers in the `hts_open_format` function I noticed the condition on  847 is identified as a blocker 
Screenshot from 2022-09-09 10-49-54 
but by the looks of it I wouldn't expect this to be a blocker. From fuzz introspector's perspective it's difficult to assess whether something is a true blocker or not. To guide the user I think it would be nice to provide the some info about the calculation/data involved on this being a blocker to enable the user to make intuitive conclusions on whether it's a true blocker or not.

Navidem:
> perhaps either perhaps mixing in with the existing blockers if some hybrid can be done or matching based on more details of the node e.g. source location.

This was a good point! Thanks @DavidKorczynski 
I update the way branch blockers are mapped to calltree nodes. Now it tries to go down the calltree on the same depth and then mark a blocker.
This how it looks like for the 3 branch blockers in `hts_hopen`
image



>  I noticed the condition on 847 is identified as a blocker

Even though such branch sides can become interesting for reaching new "execution state", but I agree it should not be listed as top blockers. The solution could be giving higher priority to the unique covered/not-covered complexities of branch sides (like the initial approach that we took) and then go for overlapping complexities.
I will address this in another PR. (#508 for tracking it)

DavidKorczynski:
> The solution could be giving higher priority to the unique covered/not-covered complexities of branch sides (like the initial approach that we took) and then go for overlapping complexities.

I think perhaps my main point is we should provide some information to the user about why the blocker is considered a blocker, to allow them to verify whether it's a true or false blocker. I think we should do this because it's not immediate obvious why the given condition is a blocker. I thought of the solution being more along the lines of providing more transparency into the calculations that made this become a blocker. We do mention it blocks X amount of complexity, but I think this might be too vague for a user trying to verify if the blocker is true or not

src/fuzz_introspector/analyses/calltree_analysis.py
```diff
@@ -170,17 +202,28 @@ def html_create_dedicated_calltree_file(
...
+            blockers_node_map = self.collect_calltree_nodes(profile.branch_blockers[:12],
+                                                            profile.function_call_depths)
+            fuzz_blocker_nodes = list(blockers_node_map.values())
+            fuzz_blocker_table = self.create_branch_blocker_table(
+                profile,
+                [],
+                "",
+                12
+            )
+        else:
```
DavidKorczynski, line 215: I think it would be nice to, down the line, to provide the option for selecting what type(s) of blocker to include

DavidKorczynski (approved):
Cool stuff!

src/fuzz_introspector/analyses/calltree_analysis.py
Navidem: Yes, that makes sense.

Navidem:
> (Can be a followup PR). Would it make sense to add this link to the call tree view itself? I.e. have a link to the blocking branch in addition to call site / function?

Also added the link to the branch blocker on the calltree:

image

PR #521, opened 2022-09-27
driver-synthesizer: refactor to avoid TypedDict

DavidKorczynski:
TypedDict causes issues when running with Python3.6 as it's only
available from Python3.8. Switching to a simple class to broaden
support.


Navidem (approved):
LGTM

PR #525, opened 2022-09-30
oss-fuzz: enable downloading public OSS-Fuzz corpus

DavidKorczynski:
Downloads the public OSS-Fuzz corpus for a project in coverage or introspector modes when --download-public-corpus is passed.

Fixes: https://github.com/ossf/fuzz-introspector/issues/452

### DIFF
--- a/oss_fuzz_integration/runner.py
+++ b/oss_fuzz_integration/runner.py
@@ -24,6 +24,44 @@ import shutil
 from typing import Optional
 
 
+def download_public_corpus(
+    project_name,
+    fuzzer_name,
+    target_zip
+):
+    OSS_FUZZ_PUBLIC_CORPUS = "https://storage.googleapis.com/%s-backup.clusterfuzz-external.appspot.com/corpus/libFuzzer/%s_%s/public.zip"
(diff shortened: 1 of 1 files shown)


DavidKorczynski:
Example: `python3 ../runner.py introspector libdwarf 1 --download-public-corpus`

AdamKorcz (approved):

evverx:
Thanks! I've been using this since it was merged and it works like a charm. I'm just spitballing but I wonder if it would make sense to move this option to `./infra/helper.py`? I think ideally it should be possible to build coverage reports with something like `./infra/helper.py coverage --download-public-corpora project` regardless of whether FI is used or not.

FWIW Now that FI reports can be built using the vanilla OSS-Fuzz toolchain I think I'd even try to integrate `oss_fuzz_integration/runner.py` into `./infra/helper.py` to avoid cloning the fuzz-introspector repository. I'm not sure how the OSS-Fuzz and FI repositories can be kept in sync with `runner.py` there though but it's probably because I haven't figured out how the oss-fuzz diff works.

Issue #536, opened 2022-10-05
Java Target Integration


arthurscchan:
Integrate support for Java fuzzing.

The tentative steps:

1. Add a set of java test cases with different classes, packages and library usage to demonstrate basic support for java projects.
2. Add static call graph generator tools for the post-processing. Possible choice is SOOT
https://soot-oss.github.io/soot/
3. Add additional logic to process data from java-callgraph.
4. Alteration of fuzz-introspector code to accept java call-graph structure with cross libraries and classes support
5. Consider coverage report generator for Java code project, Possible choice are modified version of JCov or JaCoco https://github.com/jacoco/jacoco / https://github.com/openjdk/jcov
6. Demonstrate full end-to-end coverage report and static call graph generation features
7. Alteration of result from sub-modules to ensure if fit the needed input data structure for fuzz-introspector's further processing.

arthurscchan:
PR #537 aim to create base java testcases for step 1.

arthurscchan:
PR #539 aim to manage testcases and add build script for the test cases. It also import necessary jar library for the built and calling to java-callgraph. 
Handling step 1 and 2 shown above.

DavidKorczynski:
In parallel with the callgraph we should look for the other program attributes fuzz introspector needs, such as information about:
- instructions in a function
- branches
- basic blocks
- source code lines
etc.

It probably manifests a bit different in comparison to C/C++ so we may need to look for alternative and similar data points.

arthurscchan:
JVM coverage report integration has been done and it has been bridged with OSS-FUZZ. Next step will be testing and bug fixing. Some minor bug fixing will be solved in Issue #629.

DavidKorczynski:
Well done getting this through @arthurscchan -- super excited to see this.

AdamKorcz:
Awesome stuff!

PR #537, opened 2022-10-05
tests: Add java testcases

arthurscchan:
Create java-based testcases

For step 1 of Java target integration issue #536


tests/java/test1/TestFuzzer.java
```diff
@@ -0,0 +1,23 @@
+// Copyright 2022 Google LLC
```
DavidKorczynski, line 1: Fix copyright for fuzz introspector authors

DavidKorczynski (changes requested):

tests/java/test1/TestFuzzer.java
arthurscchan: Fixed

tests/java/test5/README.md
```diff
@@ -0,0 +1 @@
+Test case with reachable and unreachable methods in different class in different
```
DavidKorczynski, line 1: missing text? What's the difference between this and test 4?

DavidKorczynski (changes requested):

tests/java/test5/README.md
arthurscchan: yes missing text.
test4 and text5 both have method in different class located in different java file, but test4 are packed in the same package as TestFuzzer while test5 are not. Thus test5 will need additional import statement.

DavidKorczynski (approved):

PR #543, opened 2022-10-12
Fix Call Graph Generation to match fuzz-introspector need.

arthurscchan:
This PR is targeting step 2 and 3 from Issue https://github.com/ossf/fuzz-introspector/issues/536


arthurscchan:
This PR is old trial test for Wala, could be merged or could be ignored since we are not using wala now. Merged to keep record of earlier work only.

arthurscchan:
This PR is old trial test for Wala, closing it and keep in local repository for reference in the future.

PR #550, opened 2022-10-18
Add case study for bzip2

swirsz:
Signed-off-by: Steven Wirsz <swirsz@gmail.com>

### DIFF
--- a/doc/CaseStudies.md
+++ b/doc/CaseStudies.md
@@ -84,3 +84,26 @@ Looking at the [Fuzz blockers](https://storage.googleapis.com/oss-fuzz-introspec
 the culprit was that `file` needs to be passed data via a proper file descriptor
 to exercise its elf-related codepath, and thus a file-based fuzzer was [promptly added](https://github.com/google/oss-fuzz/pull/8542),
 [bumping the coverage close to 90%]( https://storage.googleapis.com/oss-fuzz-introspector/file/inspector-report/20220930/fuzz_report.html ).
+
+# [bzip2](https://storage.googleapis.com/oss-fuzz-introspector/bzip2/inspector-report/20221103/fuzz_report.html)
+
+Introspector report: [link](https://storage.googleapis.com/oss-fuzz-introspector/bzip2/inspector-report/20221016/fuzz_report.html)
+
+[Bzip2's fuzz drivers hit 60%](https://storage.googleapis.com/oss-fuzz-introspector/bzip2/inspector-report/20221016/fuzz_report.html) of the functions of the project by calling the compression and decompression functions with 2 fuzz drivers. The [calltrees of bzip2_compress_target](https://storage.googleapis.com/oss-fuzz-introspector/bzip2/inspector-report/20221015/calltree_view_0.html) and [bzip2_decompress_target](https://storage.googleapis.com/oss-fuzz-introspector/bzip2/inspector-report/20221015/calltree_view_1.html) do not show any significant fuzz blockers. 
+
+The file handling code is not being addressed, as seen in the [Project functions  overview](https://storage.googleapis.com/oss-fuzz-introspector/bzip2/inspector-report/20221015/fuzz_report.html#Project-functions-overview) and by the suggestions in the remaining optimal interesting functions.
+
+![functions](https://user-images.githubusercontent.com/5808510/196543962-b65e4197-6f74-4d2c-b6f9-e12d3a9be034.png)
+
+Based on these hints, a [single new file-based fuzzer](https://github.com/google/oss-fuzz/pull/8790) was added to hit [these recommended functions.](https://github.com/libarchive/bzip2/blob/master/bzlib.c)  The resulting fuzzer creates a temporary file, calls BZ2_bzdopen, then BZ2_bzwrite, followed by BZ2_bzread to read what was just written, then BZ2_bzclose, hitting all of the optimal interesting functions.
+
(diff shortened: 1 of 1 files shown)


Navidem (approved):
Looks good! Thanks for the interesting addition.

doc/CaseStudies.md
```diff
@@ -84,3 +84,19 @@ Looking at the [Fuzz blockers](https://storage.googleapis.com/oss-fuzz-introspec
...
+Bzip2's current fuzzers call the compression and decompression functions, but the file handling code is not being addressed.  This is seen
+by the suggestions in the remaining optimal interesting functions.
+
+![interest](https://user-images.githubusercontent.com/5808510/196543962-b65e4197-6f74-4d2c-b6f9-e12d3a9be034.png)
+
+Based on these hints, a [new file-based fuzzer](https://github.com/google/oss-fuzz/pull/8790) was added to hit [these recommended functions.](https://github.com/libarchive/bzip2/blob/master/bzlib.c)  Only a single fuzzer was needed to create a temporary file, open it via a file descriptor, then write to and then read from this file.
+
+![calltree](https://user-images.githubusercontent.com/5808510/196543332-4c0554ac-43e9-4c16-8452-df79524608a8.png)
+
+The addition of this fuzzer increases the project's coverage to 80%.
```
Navidem, line 101: We can wait another day to have the new reports generated and link it here.

Navidem: Here we go: https://storage.googleapis.com/oss-fuzz-introspector/bzip2/inspector-report/20221019/fuzz_report.html

Please use this report to provide more information on the improvements.

DavidKorczynski:
Nice work @swirsz -- I wonder what's missing in order to achieve 100% reachability of the project, which would be neat. 

Looking at one of the recent reports in the "Project functions overview" table, and sorting by the rightmost column, "Undiscovered complexity" I get:
Screenshot from 2022-10-21 10-33-02

It looks like the `BZ2` family of functions could be hit (https://storage.googleapis.com/oss-fuzz-coverage/bzip2/reports/20221020/linux/src/bzip2/bzlib.c.html#L1472). It would be neat if you made a fuzzer for those targets as that would get reachability to 100% -- the first of it's kind in the fuzz introspector world I believe! Coverage will probably stay a bit lower because of e.g. error-handling code (we could deploy some mechanism that forces some of the routines that trigger error handling to fail in order to get 100% coverage, e.g. force malloc to fail once in a while).

Note the picture also shows a bug -- some missing coverage interpretation for `LLVMFuzzerTestOneInput` and some third-party functions!

swirsz:
@DavidKorczynski Very interesting challenge. I submitted another [PR](https://github.com/google/oss-fuzz/pull/8858) to try to reach that 100%.  FI reports 95% coverage (85 out of 89) and 99% CC coverage.

missed

Note that BZ2_bzdopen is fully utilized by one of the fuzz drivers.

DavidKorczynski:
That's very cool indeed @swirsz -- I'll see if I can sort these false positives in terms of coverage soon -- at least LLVMFuzzerTestOneInput should be covered. Do you know if the `fprintf` and `__assert_fail` are indeed reached?

swirsz:
There is no __assert_fail function, and FI's links to both of those references are invalid. There is a BZ2_bz__AssertH__fail function. 

I tried to make an arbitrary call to BZ2_bz__AssertH__fail and make an arbitrary fprintf statement (shows up green in FI's call tree) but the coverage still shows 85/89 functions covered and fprintf uncalled. (Same coverage as before manually adding these 2 statements)

DavidKorczynski:
Thanks for the details @swirsz -- I'll look at this in a few days and then I'll sort this out in terms of ensuring the functions that are not there should not be in the table.

DavidKorczynski:
@swirsz I think I got this one identified/solved. 

It's because of https://github.com/ossf/fuzz-introspector/blob/a47cb284d1f65cd51db809e46d96179a26497bb9/src/fuzz_introspector/constants.py#L44

which is used here: 
https://github.com/ossf/fuzz-introspector/blob/737ba7226934acdc724f4d81daa1bcb9f5b14be6/src/fuzz_introspector/utils.py#L247-L257

and thus blocks the functions in the list, which are used when finding "functions reached" for a given function:

https://github.com/ossf/fuzz-introspector/blob/737ba7226934acdc724f4d81daa1bcb9f5b14be6/src/fuzz_introspector/datatypes/function_profile.py#L47

Uncommenting the `assert` and `printf` logic from the banned list, I get:
Screenshot from 2022-10-31 21-47-02

DavidKorczynski:
@swirsz after unblocking these things, bzip2 gets:

Screenshot from 2022-10-31 22-49-04


https://github.com/ossf/fuzz-introspector/pull/577 fixes this!

DavidKorczynski:
@swirsz @Navidem 100% reachability here: https://storage.googleapis.com/oss-fuzz-introspector/bzip2/inspector-report/20221103/fuzz_report.html

Does it make sense to add this to the doc and then let's get this merged?

Navidem:
> https://storage.googleapis.com/oss-fuzz-introspector/bzip2/inspector-report/20221103/fuzz_report.html

This is amazing! Thanks @swirsz and @DavidKorczynski.

> Does it make sense to add this to the doc and then let's get this merged?

Definitely we should use the latest report, but AFAIK Steven wanted to update the posting text with more details or something. So I'd wait for him to make another push.

Navidem:
@swirsz we just need changes in `doc/CaseStudies.md`.

swirsz:
cleaned up the PR

Navidem:
Thanks!

Issue #195, opened 2022-04-08
use caching in compiler plugin to reduce resources used


DavidKorczynski:
Many projects on OSS-Fuzz are killed when we are collecting data about all of the functions. We can improve the performance in this area a lot as it currently duplicates a ton of work for many functions.

evverx:
FWIW FI appears to get killed by the OOM killer locally sometimes as well:
```
Aug 26 14:23:40 H kernel: Out of memory: Killed process 411425 (ld.gold) total-vm:1692920kB, anon-rss:1613104kB, file-rss:0kB, shmem-rss:0kB, UID:1000 pgtables:3344kB oom_score_adj:0
Aug 26 14:23:40 H systemd[1]: user@1000.service: A process of this unit has been killed by the OOM killer.
Aug 26 14:23:40 H systemd[736]: libpod-498855f30a8282abcc71721286f426b0310b43a5642bd63086619f2941e0c473.scope: A process of this unit has been killed by the OOM killer.
Aug 26 14:23:41 H admiring_benz[409266]: [393/400] clang++  -o fuzz-dhcp-client fuzz-dhcp-client.p/src_libsystemd-network_fuzz-dhcp-client.c.o -L/usr/lib/clang/15.0.0/lib/linux -Wl,--as-needed -Wl,--allo>
Aug 26 14:23:41 H admiring_benz[409266]: FAILED: fuzz-dhcp-client
Aug 26 14:23:41 H admiring_benz[409266]: clang++  -o fuzz-dhcp-client fuzz-dhcp-client.p/src_libsystemd-network_fuzz-dhcp-client.c.o -L/usr/lib/clang/15.0.0/lib/linux -Wl,--as-needed -Wl,--allow-shlib-un>
Aug 26 14:23:41 H admiring_benz[409266]: We are now in the FuzzIntrospector module pass
Aug 26 14:23:41 H admiring_benz[409266]: [Log level 1] : 14:20:48 : Fuzz introspector is running
Aug 26 14:23:41 H admiring_benz[409266]: [Log level 1] : 14:20:48 : Running introspector on ld-temp.o
Aug 26 14:23:41 H admiring_benz[409266]: [Log level 1] : 14:20:48 : This is a fuzzer, performing analysis
Aug 26 14:23:41 H admiring_benz[409266]: [Log level 2] : 14:20:48 : Using default configuration
[...]

Issue #322, opened 2022-06-10
[OSS-Fuzz] Projects fail even after using old pass manager


Navidem:
There are the projects still failing:
```
Cyclonedds
dart
git
gnutls
keystone
libcoap
libfido2
libspectre
libvips
nss
opencv
opensc
wolfssl
```

They were being successfully built before https://github.com/google/oss-fuzz/pull/7788 merged, and https://github.com/google/oss-fuzz/pull/7828 did not help to fix them (failing with different kinds of error now).

oliverchang:
@AdamKorcz @DavidKorczynski thoughts?

Some snippets:

https://oss-fuzz-build-logs.storage.googleapis.com/log-f76d4e66-6430-4815-8fd9-8a93fc13aa70.txt
```
Step #7 - "compile-libfuzzer-introspector-x86_64": make: *** [Makefile:3588: fuzz-pack-idx] Error 1
Step #7 - "compile-libfuzzer-introspector-x86_64": make: *** Waiting for unfinished jobs....
Step #7 - "compile-libfuzzer-introspector-x86_64": /tmp/lto-llvm-e3f060.o:ld-temp.o:function parse_archive_args: error: undefined reference to 'common_exit'
Step #7 - "compile-libfuzzer-introspector-x86_64": /tmp/lto-llvm-e3f060.o:ld-temp.o:function create_branches_recursively: error: undefined reference to 'common_exit'
Step #7 - "compile-libfuzzer-introspector-x86_64": /tmp/lto-llvm-e3f060.o:ld-temp.o:function dwim_branch_start: error: undefined reference to 'common_exit'
Step #7 - "compile-libfuzzer-introspector-x86_64": /tmp/lto-llvm-e3f060.o:ld-temp.o:function setup_tracking: error: undefined reference to 'common_exit'
Step #7 - "compile-libfuzzer-introspector-x86_64": clang-14: error: linker command failed with exit code 1 (use -v to see invocation)
Step #7 - "compile-libfuzzer-introspector-x86_64": make: *** [Makefile:3588: fuzz-pack-headers] Error 1
Step #7 - "compile-libfuzzer-introspector-x86_64": /tmp/lto-llvm-4ec2f3.o:ld-temp.o:function parse_archive_args: error: undefined reference to 'common_exit'
[...]

PR #567, opened 2022-10-25
llvm: write all functions without storing all in memory and delay function reachability for each function to post processing

DavidKorczynski:
This is in contrast to keeping it all in memory during the compilation. This will avoid significant memory overhead for some projects. Also avoid iterating the calltree for each function in the llvm module since we only use this to capture the functions reachable as well as function calldepth. Instead, we just capture outgoing edges for each function, and then propagate this in the python logic. This reduces the overhead of fuzz introspector because the llvm operations are resource heavy. Memory footprint is reduced a lot so we should be able to see more builds compile (e.g. the ones that are killed in the referenced issue), plus, it will speed up compilation by a lot.

Ref: https://github.com/ossf/fuzz-introspector/issues/195

### DIFF
--- a/frontends/llvm/lib/Transforms/FuzzIntrospector/FuzzIntrospector.cpp
+++ b/frontends/llvm/lib/Transforms/FuzzIntrospector/FuzzIntrospector.cpp
@@ -277,7 +277,8 @@ struct FuzzIntrospector : public ModulePass {
   StringRef removeDecSuffixFromName(StringRef funcName);
   std::string getNextLogFile();
   bool shouldRunIntrospector(Module &M);
-  FuzzerFunctionList wrapAllFunctions(Module &M);
+
+//  FuzzerFunctionList wrapAllFunctions(Module &M);
   std::string getFunctionFilename(Function *F);
   int getFunctionLinenumber(Function *F);
   std::string resolveTypeName(Type *t);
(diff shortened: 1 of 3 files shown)


DavidKorczynski:
This will affect various builds, e.g. some from https://github.com/ossf/fuzz-introspector/issues/322#issuecomment-1157866448

DavidKorczynski:
https://github.com/ossf/fuzz-introspector/blob/main/oss_fuzz_integration/test_projects.py passes

Navidem:
> Also avoid iterating the calltree for each function in the llvm module

It is not clear to how the post-process would know about all functions if llvm module does not profile them?

DavidKorczynski:
> It is not clear to me how the post-process would know about all functions if llvm module does not profile them?

The llvm module profiles each function, but some of the logic is moved to the post-processing where it's cheaper. The "functions reached" and "function call depth" for each function is currently set by going through the calltree of each function using `extractCalltree` in the llvm module. `extractCalltree` is very expensive and does a lot of double-work when applied on all functions in the module as it iteratively goes through the calltree. Instead, we continue to call it in `wrapAllFunctions` but avoid doing it recursively, i.e. call `extractCalltree` to only collect outgoing edges for each function. In the post-processing we then generate all functions reached and function call depth by iteratively going through the outgoing edges for each function. The post-processing logic is done by the added `_propagate_functions_reached` which is called by `accummulate_profile` function. In essence, we do the same thing but in a much cheaper way (simple string set/list operations versus all the llvm IR instruction handling that happens in `extractCalltree`).

Navidem:
Thanks for clarification! Sounds great to speed up llvm plugin.

Navidem (approved):
LGTM

PR #570, opened 2022-10-26
[Java-Integration] Extend soot CFG generation for oss-fuzz project

arthurscchan:
Related to step 4~5 for issue #536 

One of the potential problem is maven build ignoring some of the source and debugging dependencies during compiling and packaging, thus additional options needed to be fine tune for the oss-fuzz java project building in order to leave source trace and debugging mark for soot to pick up for static analysis.

The ultimate goal is to include call graph and extra information for static analysis of the whole project


arthurscchan:
Also related to issue #571
Considering multiple jar file, how it should be used to distinguish separate entry points. Also providing example for this.

PR #588, opened 2022-11-07
[Java-Integration]: Match frontends

arthurscchan:
Step 5~6 of issue #536 

Trying to match the output of soot with the frontends code of fuzz-introspector.

### DIFF
--- a/frontends/java/soot/src/main/java/ossf/fuzz/introspector/soot/CallGraphGenerator.java
+++ b/frontends/java/soot/src/main/java/ossf/fuzz/introspector/soot/CallGraphGenerator.java
@@ -359,6 +359,8 @@ class CustomSenceTransformer extends SceneTransformer {
 			if (!merged) {
 				className = method.getDeclaringClass().getName();
 			}
+		} else {
+			className = method.getDeclaringClass().getName();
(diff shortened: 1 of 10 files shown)


frontends/java/soot/src/main/java/ossf/fuzz/introspector/soot/CallGraphGenerator.java
```diff
@@ -280,16 +283,24 @@ protected void internalTransform(String phaseName, Map<String, String> options)
...
-				System.out.println(om.writeValueAsString(config) + "\n");
-			} catch (JsonProcessingException e) {
-				e.printStackTrace();
-			}
+		try {
+			File file = new File("fuzzerLogFile-" + this.entryClassStr + ".data");
+			file.createNewFile();
+			FileWriter fw = new FileWriter(file);
+			fw.write(extractCallTree(callGraph, this.entryMethod, 0, -1));
+		    fw.close();
```
DavidKorczynski, line 291: some formatting/indentation issues? as this code matures we should start to adopt some more stringent linting -- feel free to add some linting checks in the CI as well when you're ready.

PR #589, opened 2022-11-07
frontends: python: scan file tree instead of one up

DavidKorczynski:
This enables more data to be included in the static analysis, which is needed to produce reports for google cloud projects amongst others.


AdamKorcz (approved):

Navidem (approved):
LGTM

PR #603, opened 2022-11-09
[Java-Integration]: Fix bug

arthurscchan:
Fix bug left in PR #588 
Related to issue #536


DavidKorczynski (changes requested):
Could you clarify what the bug is and how this fixes it?

arthurscchan:
When all element subtree has been ignored during the process, it will create an empty subtree, but there is a bug which have missed checking to see if the subtree is empty. This create an empty call tree elements which is not expected and cause the html report generation reading into the first elements of am empty list and create and index out of bound exception. Fixing the code to skip the tree generation if all elements are ignored. Also fixed some duplication of exception handling in the code and missed exclude list checking.

DavidKorczynski (approved):

Issue #598, opened 2022-11-09
per fuzzer coverage links broken in util-linux


DavidKorczynski:
Ref: https://github.com/util-linux/util-linux/pull/1874#issuecomment-1308622017

On the util-linux project, the "per-fuzzer-coverage" links are broken:

We get this link: https://storage.googleapis.com/oss-fuzz-coverage/util-test_mount_fuzz/linux/reports-by-target/20221109/test_mount_fuzz/linux/report.html

Where the correct link is: https://storage.googleapis.com/oss-fuzz-coverage/util-linux/reports-by-target/20221109/test_mount_fuzz/linux/report.html

DavidKorczynski:
Looks like it's because this substitution is too strong: https://github.com/ossf/fuzz-introspector/blob/28feb560444a8c94177a7bc558d9209fec2bcc08/src/fuzz_introspector/utils.py#L240-L244

DavidKorczynski:
`.replace("/linux/", f"/{target_name}/linux/")` should fix this

evverx:
I seem to have overlooked this. Thanks for noticing and fixing this!

DavidKorczynski:
The fix for this was wrong:
There is no `/` after `linux` in the coverage url from OSS-Fuzz: https://github.com/google/oss-fuzz/blob/8b9cf2d081f5ae8d57847b9b20308dad645e8788/infra/build/functions/build_and_run_coverage.py#L306-L307

PR #605, opened 2022-11-10
tests: add url per-target testcase

DavidKorczynski:
Ref: https://github.com/ossf/fuzz-introspector/issues/598


DavidKorczynski:
This overwrites a previous fix that was wrong: https://github.com/ossf/fuzz-introspector/issues/598#issuecomment-1310256778

PR #604, opened 2022-11-10
[Java-Integration]: Create Jvm branch for fuzz-introspector html report

arthurscchan:
Create jvm branch for handling jvm based project on fuzz-introspector html report generation.

Related to issue #536

PR #607, opened 2022-11-14
[Java-integration]: Fix Callgraph display

arthurscchan:
Fix error logic to allow proper display of call graph for JVM branch
Related to issue #536 
Fixing bugs from PR #604

### DIFF
--- a/src/fuzz_introspector/datatypes/fuzzer_profile.py
+++ b/src/fuzz_introspector/datatypes/fuzzer_profile.py
@@ -99,6 +99,7 @@ class FuzzerProfile:
             return os.path.basename(self.fuzzer_source_file).replace(".py", "")
 
         elif self._target_lang == "jvm":
+            # TODO Handle jvm fuzzer source file
             logger.info("TODO Handle jvm fuzzer source file")
 
         return self.fuzzer_source_file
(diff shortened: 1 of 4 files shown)


src/fuzz_introspector/datatypes/fuzzer_profile.py
```diff
@@ -438,6 +438,11 @@ def _load_coverage(self, target_folder: str) -> None:
                     self.all_class_functions
                 )
         elif self.target_lang == "jvm":
+            self.coverage = code_coverage.load_llvm_coverage(
```
DavidKorczynski, line 441: This change looks a bit strange to met: if the target is `jvm` then use `load_llvm_coverage`? LLVM coverage is only for C/C++, or is it also for java?

arthurscchan: Because we does not have coverage report yet for java, so this is a temporary measure to use the logic of the cpp one to build up an empty coverage report to show.

DavidKorczynski (approved):

DavidKorczynski:
@arthurscchan I had to resolve a conflict when merging this, but am not sure I did the right thing. Could you review https://github.com/ossf/fuzz-introspector/pull/607/commits/0b38f6a6f1dd60742c56063767f9a07bc107aa7b and submit a patch if necessary

PR #611, opened 2022-11-14
[Java Integration]: Fix jvm data bug

arthurscchan:
Fix the bug that some method and data are missing from the report. 
Also, now the call tree are displaying method arguments, to distinguish same name method with different argument for unique identification purpose.

Related to issue #536 
Aim to fix bug for PR #604

### DIFF
--- a/src/fuzz_introspector/datatypes/fuzzer_profile.py
+++ b/src/fuzz_introspector/datatypes/fuzzer_profile.py
@@ -86,7 +86,9 @@ class FuzzerProfile:
         elif self.target_lang == "python":
             return self.entrypoint_fun
         elif self.target_lang == "jvm":
-            return "fuzzerTestOneInput"
+            return "fuzzerTestOneInput(com.code_intelligence.jazzer.api.FuzzedDataProvider)"
+        else:
+            return None
 
     @property
     def identifier(self):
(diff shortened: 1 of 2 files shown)


arthurscchan:
Because the report group functions with same name from the same source file as one row, it generates false cases for java because in the same source file (class) there could be more than one function with the same name but different signature and arguments. That original logic does not handle this cases. In order to accommodate this cases, the new logic consider the whole method signature (including name and arguments) as the method name in order to give the method an unique identifiers for the further process in the report. It also give better understanding of which method is being called and covered by the code.
Sample of the report are as follows.
Screenshot 2022-11-14 at 9 13 13 PM
Screenshot 2022-11-14 at 9 13 05 PM
Screenshot 2022-11-14 at 9 12 58 PM
Screenshot 2022-11-14 at 9 13 24 PM
Screenshot 2022-11-14 at 9 13 33 PM

src/fuzz_introspector/datatypes/fuzzer_profile.py
```diff
@@ -100,22 +102,20 @@ def identifier(self):
...
+            pass
 
         return self.fuzzer_source_file
 
     def has_entry_point(self) -> bool:
         """Returns whether an entrypoint is identified"""
         if self.target_lang == "c-cpp":
-            if "LLVMFuzzerTestOneInput" in self.all_class_functions:
-                return True
+            return self.entrypoint_function in self.all_class_functions
```
DavidKorczynski, line 112: Is `self.entrypoint_function` guaranteed to be `LLVMFuzzerTestOneInput` for `c-cpp` targets?

arthurscchan: no, but this is a property of the class defined in Line 85. I just to point back to that, to manage a better code style to only need to change one location.

PR #621, opened 2022-11-22
[Java Integration] Connect oss-fuzz jvm project with fuzz-introspector runner

arthurscchan:
Add logic in fuzz-introspector runner to include all necessary steps (Build, run, generating coverage, generating html report) for oss-fuzz jvm project, thus connecting them to fuzz introspector.

This PR consist of the following steps for JVM project connection:

1. Apply patch for fuzz-introspector logic into the jvm project's build.sh before building the project.
2. Build the project normally, run the fuzzers, rebuild the project with coverage sanitizer to generate coverage report
3. Goto correct build result folder, set the correct permission.
4. Run fuzz-introspector main logic to generate HTML report
5. Launch the python http.server in the directory to showcase the HTML report.

This is related to the last step of issue #536

### DIFF
--- a/oss_fuzz_integration/jvm.patch
+++ b/oss_fuzz_integration/jvm.patch
@@ -37,11 +37,13 @@ class=${class:1}
 cd $OUT
 rm -rf fuzz-introspector
 git clone https://github.com/arthurscchan/fuzz-introspector
+# Require change to "cd fuzz-introspector/frontends/java"
 cd fuzz-introspector/frontends/java/soot
-git checkout coverage-report
 sed -i 's/mvn/$MVN/g' run.sh
 ./run.sh -j $jarfile -c $class -m fuzzerTestOneInput
 cd $OUT
+# Require change to "cp fuzz-introspector/frontends/java/*.data ./"
 cp fuzz-introspector/frontends/java/soot/*.data ./
+# Require change to "cp fuzz-introspector/frontends/java/*.data.yaml ./"
 cp fuzz-introspector/frontends/java/soot/*.data.yaml ./
 rm -rf fuzz-introspector
--- a/oss_fuzz_integration/jvm.patch
+++ b/oss_fuzz_integration/jvm.patch
@@ -1,0 +1,47 @@
+# Packing for fuzz-introspector
+jarfile=
+class=
+
+for fuzzer in $(find $SRC -name '*Fuzzer.java'); do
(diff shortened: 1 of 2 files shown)


oss_fuzz_integration/runner.py
```diff
@@ -62,6 +65,30 @@ def download_full_public_corpus(project_name, target_corpus_dir: None):
         subprocess.check_call(f"unzip {target_zip} -d {target_fuzzer_dir}/", shell=True)
 
 
+def prepare_introspector(curr_dir):
```
DavidKorczynski, line 68: I think we should avoid this. Is there a reason for not putting it in the diff and handling it like we do with other languages?

oss_fuzz_integration/jvm.patch
```diff
@@ -0,0 +1,47 @@
+# Packing for fuzz-introspector
```
DavidKorczynski, line 1: I'd prefer to put this logic into the OSS-Fuzz environment rather, i.e. having it as a script we run in the base-builder

oss_fuzz_integration/runner.py
```diff
@@ -305,37 +412,70 @@ def introspector_run(
...
+
+        server_directory = os.path.abspath(latest_corpus_dir)
+    else:
+        # Build sanitizers with introspector
+        build_project(project_name, sanitizer="introspector")
+
+        # get the latest corpus
+        latest_corpus_dir = get_recent_corpus_dir()
+
+        # copy over inpsoector and coverage reports
```
DavidKorczynski, line 451: typo

oss_fuzz_integration/runner.py
```diff
@@ -305,37 +412,70 @@ def introspector_run(
...
-    
-    # Build sanitizers with introspector
-    build_project(project_name, sanitizer="introspector") 
 
-    # get the latest corpus
-    latest_corpus_dir = get_recent_corpus_dir()
+    curr_dir = os.path.abspath(".")
 
-    # copy over inpsoector and coverage reports
+    if get_project_lang(project_name) == 'jvm':
```
DavidKorczynski, line 418: I'd prefer to put the language specific items into the oss-fuzz environment as much as we can. Can the logic that follows be put in the base-builder?

DavidKorczynski (changes requested):
Very nice, am excited about this! My main gripe with the current set up is that we should get it closer into the oss-fuzz environment similar. The logic for handling jvm reports should be put in the base-builder. I'd prefer we do that from the get-go by way of our oss-fuzz-patches.diff file

PR #627, opened 2022-11-23
oss-fuzz: refine jvm set up

DavidKorczynski:
- Change the paths where indicated by the `# Require change to "cp fuzz-introspector/frontends/java/*.data ./"
- Modify the `runner.py` to find the `jvm.patch` relative to the `runner.py` file instead of the current working dir
- wrap the `jvm.patch` script in a condition that's only true once, meaning we can run `runner.py` a number of times. This is a quick hack for a temporary situation.
- fix a bug in `runner.py` for correctly identifying fuzz targets.

### DIFF
--- a/oss_fuzz_integration/jvm.patch
+++ b/oss_fuzz_integration/jvm.patch
@@ -36,14 +36,12 @@ class=${class:1}
 
 cd $OUT
 rm -rf fuzz-introspector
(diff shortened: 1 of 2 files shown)


DavidKorczynski:
CC @arthurscchan

arthurscchan:
lgtm

arthurscchan:
Related to summarise connection to oss-fuzz for issue #536

Issue #629, opened 2022-11-24
jvm issues


DavidKorczynski:
Umbrella issue for minor jvm issues

- runtime coverage functions is above reachable functions
- urls is missing some parts (apache-commons-cli is an example), including .java

DavidKorczynski:
Example wrong url:
http://localhost:8008/covreport/linux/org/apache/commons/cli/DefaultParser.html#L530

Right link: http://localhost:8008/covreport/linux/org.apache.commons.cli/DefaultParser.java.html#L530

arthurscchan:
This issue is a on-going bug fixing issue for Issue #536.

arthurscchan:
> Example wrong url: http://localhost:8008/covreport/linux/org/apache/commons/cli/DefaultParser.html#L530
> 
> Right link: http://localhost:8008/covreport/linux/org.apache.commons.cli/DefaultParser.java.html#L530

The possible reason is Jacoco report does not care non-instruction statements, will need to check if there is option in the jacoco report to allow inspection of those statements and fix it on oss-fuzz side.

arthurscchan:
Another possible issue is the difference in runtime coverage and static reachability analysis. Some methods are missing in the reachability analysis making runtime coverage not able to handle those methods correctly.

arthurscchan:
Additional figures that are needed to modify is the method reached for each fuzzers, instead of just counting the first layer of reachable methods, count deep into all layer of method reached and add them and show a total sum of all reachable methods from the whole call tree.

arthurscchan:
Unit testing for the frontend java code needed to be added.

arthurscchan:
PR #632 and #634 aims to handle the link resolve bugs happened for linking the callgraph functions to the jacoco source report. It includes the missing of .java for code handling, mishandled link for package folder and mishandled link for the source code reference of subclasses.

arthurscchan:
Further unit testing are being monitored in Issue #630

PR #633, opened 2022-11-25
[JVM issue] Add code to patch line number of jacoco source report

arthurscchan:
Solving issue #629


DavidKorczynski (approved):
Very nice with a fix to make the jacoco reports more linkable

PR #635, opened 2022-11-28
[Java-issue] Fix jacoco report patching logic and timing

arthurscchan:
Solving invalid line number reference to jacoco report, mentioned in issue #629.
Solving the patch code in issue #633, the patch logic is basically correct, but are called and executed in wrong location before the report even exists in the designated location. Change the code to patch the html report files after they have been moved to the designated location.


src/fuzz_introspector/datatypes/fuzzer_profile.py
```diff
@@ -454,9 +454,6 @@ def _load_coverage(self, target_folder: str) -> None:
                 self.coverage.correlate_jvm_method_with_coverage(
                     self.all_class_functions
                 )
-
```
DavidKorczynski, line 457: The changes in this file reverse what we just pushed here: https://github.com/ossf/fuzz-introspector/pull/634

I assume this should not be the case?

DavidKorczynski (changes requested):

src/fuzz_introspector/datatypes/fuzzer_profile.py
arthurscchan: Yes, you are right, I also just found that out in my testing and verification and then I see your message.

oss_fuzz_integration/runner.py
```diff
@@ -105,6 +105,45 @@ def patch_jvm_build(project_build_path):
             file_handle.write(content)
 
 
+def patch_jvm_source_report(server_directory):
+    """
+    Jacoco HTML report showing the source coverage does not have
```
DavidKorczynski, line 110: Very nice fuction-level comment

DavidKorczynski (approved):

PR #637, opened 2022-11-28
Allow pointing runner to local source directories

evverx:
There are use cases where it would be desirable to be able to point FI to specific local directories when projects are built. For example systemd can't be built with FI because it links its fuzz targets against shared libraries and any patches getting FI to work are unlikely to be accepted upstream. That shouldn't prevent anyone from using FI with it though as long as their local forks are compatible with FI. Also when fuzz targets and/or code are changed or when PRs are analyzed locally it should be possible to confirm that things have gotten better or at least haven't gotten worse.

It's probably worth mentioning that since `./infra/helper.py` doesn't clean up after itself for this to work build scripts should clean up their build directories themselves using `make clean`, `git clean -dxf`
 `rm ./build` or something like that. That looks like an
`infra/helper.py` bug and should probably be fixed there eventually by switching to ephemeral mounts where whatever is written by build scripts gets deleted once containers are stopped.

### DIFF
--- a/oss_fuzz_integration/runner.py
+++ b/oss_fuzz_integration/runner.py
@@ -75,6 +75,7 @@ def download_full_public_corpus(project_name, target_corpus_dir: None):
 
 def build_project(
     project_name,
+    source_dir = None,
     sanitizer = None,
     to_clean = False
 ):
@@ -86,6 +87,8 @@ def build_project(
     if to_clean:
         cmd.append("--clean")
     cmd.append(project_name)
+    if source_dir is not None:
+        cmd.append(source_dir)
 
     try:
         subprocess.check_call(" ".join(cmd), shell=True)
@@ -371,6 +374,7 @@ def setup_next_corpus_dir(project_name):
 
 def complete_coverage_check(
     project_name: str,
+    source_dir: Optional[str],
     fuzztime: int,
     job_count: int,
     corpus_dir: Optional[str],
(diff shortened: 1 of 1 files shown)


DavidKorczynski (approved):
lgtm

To confirm, the issue you mention with regards to shared libraries is it the same as https://github.com/google/oss-fuzz/issues/7598 ?

evverx:
> is it the same as https://github.com/google/oss-fuzz/issues/7598 ?

I totally forgot about that issue. Yes it is. Since systemd links its fuzz targets against shared libraries the LTO pass misses them and as a result the reports are somewhat confusing because a large chunk of the code is missing. So systemd can actually be built by applying a meson patch like https://github.com/google/oss-fuzz/pull/8222 but the reports aren't as useful as they can be.

Before I forget to avoid editing all the build scripts I fixed the `infra/helper.py` issue locally by applying the following patch
```sh
diff --git a/infra/helper.py b/infra/helper.py
index 8d17cb8c..caef6825 100755
--- a/infra/helper.py
+++ b/infra/helper.py
@@ -701,7 +701,7 @@ def build_fuzzers_impl(  # pylint: disable=too-many-arguments,too-many-locals,to

       command += [
           '-v',
-          '%s:%s' % (_get_absolute_path(source_path), workdir),
+          '%s:%s:O' % (_get_absolute_path(source_path), workdir),
       ]

   command += [
```
With it any changes to source directories are destroyed when containers are stopped so `helper.py` can be run a few times in a row. I'll try to send it to OSS-Fuzz once I make sure that it works with docker too. I kind of point `helper.py` to a bash script that replaces `docker` with `podman` and adds `--pull=never` to command lines when necessary.

PR #640, opened 2022-11-28
[JVM issue] Fix link bug

arthurscchan:
In optimal_targets.py (https://github.com/ossf/fuzz-introspector/blob/main/src/fuzz_introspector/analyses/optimal_targets.py#L259-L264), it uses some logic to determine the coverage source report link for a specific functions. But the logic is only suitable for c-cpp. 

In fuzzer_profile.py, there is helper functions to handle coverage source report link for c-cpp / python / jvm separately. To reduce code duplication and to allow optimal function table to use the same  coverage link resolve function, this PR move the coverage link resolve code to utils and allows both fuzzer_profile.py and optimal_targets.py to use it.

Related to issue #629 and also fixing bugs for python report.


evverx:
Looks like https://github.com/ossf/fuzz-introspector/pull/639 and https://github.com/ossf/fuzz-introspector/pull/638 were reverted when I assume the conflicts were resolved automatically during a rebase or something like that. It would be great if it was possible to bring them back.

DavidKorczynski (changes requested):
@arthurscchan please check the latest changes from fuzz-introspector are intact

arthurscchan:
Changes from PR #638 and #639 is merged in.

DavidKorczynski (approved):

PR #646, opened 2022-11-30
[JVM Issue]: Fix possible errornous calculation

arthurscchan:
Possible wrong inclusion of non-hitted function for coverage calculation.


DavidKorczynski (approved):

DavidKorczynski:
good catch

Issue #630, opened 2022-11-24
[Unit Testing] Check post processing logic base on JVM code


arthurscchan:
Add in unit testing test cases to ensure changes in the post-processing logic does not affect or break anything for the JVM processing logic.

arthurscchan:
Consider extending unit test cases in tests/ directory to cover "JVM" specific branches code to ensure they are not "accidentally" ruined.

PR #647, opened 2022-11-30
[Unit Testing]: More test cases

arthurscchan:
Add more unit testing test cases.

Related to issue #630 and extending PR #643


frontends/java/src/test/java/ossf/fuzz/introspector/soot/CustomSenceTransformerTest.java
```diff
@@ -0,0 +1,81 @@
...
+  @Test
+  public void testExcludePrefix() {
+    CustomSenceTransformer custom = new CustomSenceTransformer("", "", "abc:def:ghi");
+    assertEquals(custom.getExcludeList().size(), 3);
+    Object[] expected = {"abc", "def", "ghi"};
+    assertArrayEquals(expected, custom.getExcludeList().toArray());
+  }
+
+  @Test
+  public void sampleTestCases() throws IOException, InterruptedException {
```
DavidKorczynski, line 39: I'm not sure about this test. The problem is that it may become a somewhat blocker of progress. Down the line, if we start to extend how the output looks, then this will fail even though semantics are intact. I feel like the tests should be rooted in semantic rather than syntax -- and this one compares two outputs in a string comparison fashion.

I think the easiest approach is to make the test a bit more narrow and focus on making the tests check that certain properties hold.

WDYT?

DavidKorczynski (changes requested):

frontends/java/src/test/java/ossf/fuzz/introspector/soot/CustomSenceTransformerTest.java
arthurscchan: I agree. And new unitesting has been pushed in PR #663

arthurscchan:
Not needed anymore, using another approach in PR #663

PR #654, opened 2022-12-01
Java-Issue: Fix method naming logic error

arthurscchan:
Java method from different classes are allowed to have a same signature. This logic does not work well with fuzz-introspector which identify unique function only by function name. In order to accommodate fuzz-introspector analysing logic, changes has been made on the Soot static analysis code to include also the class name as part of the function identifiers. This PR provide a fix to the soot generation logic to include class name in the resulting .data and .data.yaml file to allow fuzz-introspector correctly handle and analyse methods from different class but with the exact same signature.

Example:
toString() methods coming from java.io.File and java.io.FileInputStream have the exact same signature, making the method name as java.io.File.toString() and java.io.FileInputStream.toString() can uniquely identify them.

Related to Issue #629 

### DIFF
--- a/frontends/java/src/main/java/ossf/fuzz/introspector/soot/CallGraphGenerator.java
+++ b/frontends/java/src/main/java/ossf/fuzz/introspector/soot/CallGraphGenerator.java
@@ -165,20 +165,16 @@ class CustomSenceTransformer extends SceneTransformer {
           continue;
         }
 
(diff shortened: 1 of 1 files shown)


DavidKorczynski (approved):
This is similar to C++ and python classes where this is also included. For example, in C++ the mangled name/demangled name holds this info, which is what we use

PR #658, opened 2022-12-02
Update contact for the code of conduct

calebbrown:
Replace the placeholder text with the correct email address as confirmed via the OpenSSF Slack.

Signed-off-by: Caleb Brown <github@calebbrown.id.au>


Navidem (approved):
Thanks!

PR #660, opened 2022-12-02
Unit Testing: Add test case for resolving coverage link

arthurscchan:
This PR aims to create unit test cases for resolving coverage link function in utils for all language, since the resolve coverage link code is moved from fuzzer_profile to utils and generalised for all supported languages.

Issue #630 

### DIFF
--- a/src/test/test_utils.py
+++ b/src/test/test_utils.py
@@ -91,3 +91,154 @@ def test_get_target_coverage_url(coverage_url: str, fuzz_target: str, res: str, 
     os.environ['FUZZ_INTROSPECTOR'] = "1"
     assert utils.get_target_coverage_url(coverage_url, fuzz_target, lang) == res
     del os.environ['FUZZ_INTROSPECTOR']
+
+@pytest.mark.parametrize(
+    ('cov_url', 'source_file', 'lineno', 'function_name', 'target_lang', 'temp_file', 'expect'),
+    [
+        (
+            'https://coverage-url.com/',
+            'fuzzlib/fuzzlib.c',
+            '13',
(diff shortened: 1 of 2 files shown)


src/test/test_utils.py
```diff
@@ -91,3 +91,154 @@ def test_get_target_coverage_url(coverage_url: str, fuzz_target: str, res: str,
...
+                 "files":{
+                     "Test":{
+                         "hash":"Test",
+                         "index":{
+                             "relative_filename":"/src/fuzz_parse.py"
+                         }
+                     }
+                 }
+            }''',
+            'https://coverage-url.com//Test.html#t13'
```
DavidKorczynski, line 157: Do we expect `...com//Test` here? I would assume we expect `...com/Test` here? Same in other tests

DavidKorczynski (changes requested):

src/test/test_utils.py
arthurscchan: Yes indeed, because both the java and python logic does add a seperator to it because the coverage link is supposed to be /covreport/linux ......

arthurscchan: this is a fail safe logic to ensure even if the coverage link does not ended with /, it still works.

DavidKorczynski: so which do we expect?

DavidKorczynski: URL sanitization to ensure there is only a single `/` is fixed in: https://github.com/ossf/fuzz-introspector/pull/661

DavidKorczynski (approved):

PR #661, opened 2022-12-02
Fix coverage link resolve logic

arthurscchan:
Fix coverage link resolve logic to ensure no double / exists in the link (except for the protocol)

### DIFF
--- a/src/fuzz_introspector/utils.py
+++ b/src/fuzz_introspector/utils.py
@@ -286,8 +286,9 @@ def resolve_coverage_link(
     target_lang: str
 ) -> str:
     """Resolves link to HTML coverage report"""
+    result = "#"
     if (target_lang == "c-cpp"):
-        return cov_url + source_file + ".html#L" + str(lineno)
+        result = os.sep + source_file + ".html#L" + str(lineno)
     elif (target_lang == "python"):
         """Resolves link to HTML coverage report for Python targets"""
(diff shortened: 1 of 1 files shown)


src/fuzz_introspector/utils.py
```diff
@@ -324,7 +324,11 @@ def resolve_coverage_link(
...
         source_file = source_file.split("$")[0]
 
-        return cov_url + os.sep + source_file + ".java.html#L" + str(lineno)
+        result = os.sep + source_file + ".java.html#L" + str(lineno)
     else:
         logger.info("Unsupported language for coverage link resolve")
-        return "#"
+
+    if result != "#":
+        result = cov_url.rstrip(os.sep) + result.replace("%s%s" % (os.sep, os.sep), os.sep)
```
DavidKorczynski, line 332: I don't think the use of `os.sep` is the way to go -- we're dealing with URLs and not filepaths. I would assume we always deal with `/` here

DavidKorczynski (changes requested):

src/fuzz_introspector/utils.py
DavidKorczynski: Also, does it make sense to make this more general? The code as it stands does not guarantee there are no double `//`:
```
>>> s1="////".replace("//","/")
>>> s1
'//'
```

DavidKorczynski (approved):

PR #663, opened 2022-12-03
Unit-Testing: Full jvm report unit testing

arthurscchan:
Use existing test cases for jvm in aid of full jvm fuzz-introspector html report generation testing.

Issue: #630


DavidKorczynski (approved):
LGTM, great stuff!

Down the line I think we should focus on making the results of the processing in `src/fuzz-introspector` to be better wrapped in appropriate data structures. This would make it possible to e.g. avoid having to parse the generated `html` file, which seem a bit brittle as it may break if we improve the generated HTML.

PR #669, opened 2022-12-06
Sink Analyser: Create new sink analyser for analysing sink functions in project

arthurscchan:
ICreate new constant to store possible function sink list of supported language.

Issue #670 


DavidKorczynski:
I think I'd prefer to have these as part of the `.py` module for the given analysis plugin. Could we store them in the module we place in https://github.com/ossf/fuzz-introspector/tree/main/src/fuzz_introspector/analyses ? Is the plan to extend https://github.com/ossf/fuzz-introspector/blob/main/src/fuzz_introspector/analyses/function_call_analyser.py or create a new module one?

arthurscchan:
I think just keep it like that temporary. Will move them when the crawler and collectio are working.

src/fuzz_introspector/analyses/sinks_analyser.py
```diff
@@ -0,0 +1,394 @@
...
+            for called_location in function_callsite_dict[fd.function_name]:
+                # Determine if this called location is covered by any fuzzers
+                fuzzer_hit = False
+                coverage = proj_profile.runtime_coverage
+                for parent_func in fd.incoming_references:
+                    try:
+                        lineno = int(called_location.split(":")[1])
+                    except ValueError:
+                        continue
+                    if coverage.is_func_lineno_hit(parent_func, lineno):
```
DavidKorczynski, line 377: Am not sure this will work for all languages and whether we should have more insights about the target language of the coverage profile, or, alternatively do some refactoring. Am happy with addressing this at a later stage.

src/fuzz_introspector/analyses/sinks_analyser.py
```diff
@@ -0,0 +1,394 @@
...
+
+        return (callsite_list, function_list)
+
+    def map_function_callsite(
+        self,
+        functions: List[function_profile.FunctionProfile],
+        callsites: List[cfg_load.CalltreeCallsite]
+    ) -> Dict[str, List[str]]:
+        """
+        This function aims to dig up the callsite for each functions
```
DavidKorczynski, line 186: nit, "each function"

src/fuzz_introspector/analyses/sinks_analyser.py
```diff
@@ -0,0 +1,394 @@
...
+        ('code.InteractiveConsole', 'interact'),
+        ('code.InteractiveConsole', 'raw_input'),
+        ('code', 'interact'),
+        ('code', 'compile_command')
+    ],
+    'jvm': [
+        ('java.lang.Runtime', 'exec'),
+        ('javax.xml.xpath.XPath', 'compile'),
+        ('javax.xml.xpath.XPath', 'evaluate'),
+        ('java.lang.Thread', 'sleep'),
```
DavidKorczynski, line 80: We should add some documentation on why each function is relevant, even just a link if it's based on a previous vulnerability or so. Am not sure if it's general knowledge to users why `java.lang.Thread.sleep` is an indication of possible code injection, but to me it's not.

DavidKorczynski (approved):
LGTM. There a few issues to look out for for am happy to merge this is in now and address them later. Most likely it will also need some assistance in order to make things work well for all frontends (untli we get a nice stable API for all frontends).

DavidKorczynski:
Please fix CI and then I'll merge

src/fuzz_introspector/analyses/sinks_analyser.py
DavidKorczynski: @arthurscchan could you please follow up on this?

Issue #623, opened 2022-11-22
Add google analytics for oss-fuzz


DavidKorczynski:
From https://github.com/google/oss-fuzz/issues/9039

Navidem:
Do we have a timeline for this?

AdamKorcz:
> Do we have a timeline for this?

By end of this week.

Navidem:
Thanks!

DavidKorczynski:
This is now fixed in https://github.com/ossf/fuzz-introspector/pull/671 -- if `G_ANALYTICS_TAG` environment variable is set with the relevant analytics tag, google analytics is enabled!

DavidKorczynski:
todo: update fuzz introspector version on OSS-Fuzz and add an environment variable with the relevant tag when building introspector builds

PR #671, opened 2022-12-07
add option for Google Analytics

AdamKorcz:
Adds the option to add Google Analytics tracking by setting the `G_ANALYTICS_TAG` env variable.

Signed-off-by: AdamKorcz <adam@adalogics.com>

### DIFF
--- a/src/fuzz_introspector/html_helpers.py
+++ b/src/fuzz_introspector/html_helpers.py
@@ -21,6 +21,7 @@ from typing import (
     Tuple,
 )
 
+import os
 from fuzz_introspector import utils
 from fuzz_introspector.datatypes import fuzzer_profile, project_profile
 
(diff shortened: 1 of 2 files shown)


src/fuzz_introspector/html_helpers.py
```diff
@@ -52,8 +53,24 @@ def html_table_add_row(elems: List[Any]) -> str:
 
 def html_get_header(calltree: bool = False,
                     title: str = "Fuzz introspector") -> str:
-    header = """<html>
+    gtag_tracking = ""
+    try:
+        gtag = os.environ['G_ANALYTICS_TAG']
+        gtag_tracking += f"""<!-- Google tag (gtag.js) -->
+                <script async src="https://www.googletagmanager.com/gtag/js?id=G-{gtag}"></script>
```
DavidKorczynski, line 60: Should we prefix the tag with `G-` or let the user set this themselves?

src/fuzz_introspector/html_helpers.py
```diff
@@ -52,8 +53,24 @@ def html_table_add_row(elems: List[Any]) -> str:
...
+                <script>
+                  window.dataLayer = window.dataLayer || [];
+                  function gtag(){{dataLayer.push(arguments);}}
+                  gtag('js', new Date());
+
+                  gtag('config', 'G-{gtag}');
```
DavidKorczynski, line 66: Should we prefix the tag with `G-` or let the user set this themselves?

src/fuzz_introspector/html_helpers.py
AdamKorcz: The prefix might not be G- AFAIK so we should not prefix it. Updated.

src/fuzz_introspector/html_helpers.py
AdamKorcz: The prefix might not be G- AFAIK so we should not prefix it. Updated.

DavidKorczynski (approved):

DavidKorczynski:
Fixes https://github.com/ossf/fuzz-introspector/issues/623

PR #676, opened 2022-12-09
JVM implementation bug: Fix jvm link with missing package name

arthurscchan:
Fix a bug that the jvm coverage links are missing because of wrong comparison and generation of java package name.


DavidKorczynski (commented):
The unit tests are failing -- is this an instance of where the test is wrong and should be updated?

DavidKorczynski:
> The unit tests are failing -- is this an instance of where the test is wrong and should be updated?

The unit tests were right -- that's great!!

DavidKorczynski (approved):

PR #680, opened 2022-12-12
jvm: improve oss-fuzz set up

DavidKorczynski:
- get rid of jvm.patch
- avoid using run.sh

Substitute the above with a simple starter script dedicated for OSS-Fuzz usage.


DavidKorczynski:
@arthurscchan this is working for `javassist` but not for `angus-mail` in that the resulting `.data` and `.data.yaml` files will only have data from the fuzzer `.jar` and not the projects' `.jar` (`angus-core.jar`). Do you see any potential issues?

arthurscchan:
There is a default exclude list 
jdk.:java.:javax.:sun.:sunw.:com.sun.:com.ibm.:com.apple.:apple.awt.
included in the execution. The logic will ignore all package start with these string (separated by ':') cause these are the default package for some native java library which are normally not necessary to be included in the project analysis. If there are some project "accidentally" used these default package name prefix, simply provide the logic with the customized set of exclude prefix excluding the one being used by the project.

arthurscchan:
> @arthurscchan this is working for `javassist`,`kryo` but not for `angus-mail` in that the resulting `.data` and `.data.yaml` files ends up only having content from the fuzzer `.jar` and not the projects' `.jar` (`angus-core.jar`). Do you see any potential issues?

So because the source file of angus-mail is using the native pakcage of com.sun which are in the default exclude list, thus those package has been ignored and so the resulting .data and .data.yaml does not contains the call in those code. This is the a possible reason for the difference.

DavidKorczynski:
After removing the exclusion prefix (including the deafult) it didn't succeed either.

Specifically, the command I run: `java -Xmx6144M -cp /fuzz-introspector/frontends/java/target/ossf.fuzz.introspector.soot-1.0.jar ossf.fuzz.introspector.soot.CallGraphGenerator angus-core.jar:BASE64EncoderStreamFuzzer.jar:ASCIIUtilityFuzzer.jar BASE64EncoderStreamFuzzer fuzzerTestOneInput rnad.`

DavidKorczynski:
The calltree for the BASE64EncoderStreamFuzzer we get is:
```
Call tree
fuzzerTestOneInput(com.code_intelligence.jazzer.api.FuzzedDataProvider) BASE64EncoderStreamFuzzer linenumber=-1
  consumeRemainingAsBytes() com.code_intelligence.jazzer.api.FuzzedDataProvider linenumber=29
  write(byte[]) com.sun.mail.util.BASE64EncoderStream linenumber=29
```

This is actually accurate in that the `write` function called by the fuzzer is just a wrapper around `write`:
https://github.com/eclipse-ee4j/angus-mail/blob/46518aad79259d12a115937894ff8f32b86a1a17/core/src/main/java/com/sun/mail/util/BASE64EncoderStream.java#L141-L144

However, the `.data.yaml` should still have more functions in them? Currently I only have two functions in `.data.yaml`

DavidKorczynski:
It seems like the `angus-core` problem is unrelated to this PR. Will open it as another issue and move on with this PR

AdamKorcz (approved):

Issue #670, opened 2022-12-06
Code Injection Sink Report: New implement


arthurscchan:
Provides report for code injection sink functions coverage and path to reach those functions if unreachable. This issue aims to develop an additional section in the fuzz-introspector report for showing additional information and suggestions on the project fuzzer ability to discover code injection in the original project. 
The report will show a list of possible code injection sink functions existing in the target project, together with its static and dynamic reachability by the given fuzzers. As they are the target for code injection, it is essential that the fuzzer does reach and hitted them during the fuzzing section to ensure it is safe guarded against code injection. If the project is not statically reachable by all provided fuzzers, it will also show suggestions on possible function candidates that the fuzzer should consider to reach the sinks.
The work needed to get this new report integration are divided into 5 parts.

Part 1: Do research and to build up a temporary list (which could be modifiable in the later stage) of sink functions for the supporting language (currently c-cpp / python / jvm).

Part 2: Go through the all functions list retrieved from the data and data.yaml files of the projects and list out the functions used that match the possible sinks list. 

Part 3: Show the reachability and fuzzer hit status of each of the shown func

Part 4: Dig into each target function and trace back to the closest callable function in the project, which allows the fuzzer to call and fuzz.

Part 5: Display the result of closable trace for those functions

Additional Task:
Automatic develop fuzzers that reach those unreached or uncovered functions

Testing and wrap up documentation

Navidem:
Thanks for creating this issue, it's a very interesting idea to employ!

> Part 1: Develop a list of sink functions

Agreed that it can be a pre-populated list of "sensitive" functions for this analysis purpose.

> Part 2: Iterate through the functions in the project (retrieved from the data and data.yaml files) and list out the functions that match the possible sinks.

If I get it right, here you will match for the function name in the list of visited functions in the call-graph.
I think we may go even further and employ a data-flow analysis here to alarm only cases that an untrusted input reaches the sensitive function. The data-flow can be from the closest covered function in the call-graph to the sensitive function.
My logic here is that, if the sensitive function is already covered, then we can report it as covered (and even provide the level of coverage as a verdict, for example low/moderate/hight). But if the sensitive function is not covered, our report shrinks to listing some sensitive functions in the code that needs further attention. Which itself is still an interesting report.

Navidem:
P.S: I should say the keyword "sink" in the title took me to the data-flow :)

arthurscchan:
[...]

PR #686, opened 2022-12-13
Sink coverage report: Commit new flow for retrieving sink coverage from analyser

arthurscchan:
Create a flow to allow the user of fuzz-introspector generate a separate json report for sink coverage information of the target project.

Issue #670

### DIFF
--- a/src/fuzz_introspector/analysis.py
+++ b/src/fuzz_introspector/analysis.py
@@ -43,7 +43,8 @@ logger = logging.getLogger(name=__name__)
 
 
 class AnalysisInterface(abc.ABC):
-    name: str
+    name: str = ""
+    json_string_result: str = ""
 
     @abc.abstractmethod
     def analysis_func(
@@ -54,8 +55,7 @@ class AnalysisInterface(abc.ABC):
         profiles: List[fuzzer_profile.FuzzerProfile],
         basefolder: str,
         coverage_url: str,
-        conclusions: List[html_helpers.HTMLConclusion],
-        json_report: bool
+        conclusions: List[html_helpers.HTMLConclusion]
     ) -> str:
         """Entrypoint for analysis instance. This function can have side effects
         on many of the arguments passed to it.
@@ -83,15 +83,9 @@ class AnalysisInterface(abc.ABC):
                            be shown at the top of the report page.
         :type conclusions: List[html_helpers.HTMLConclusion]
 
-        :param json_report: A bool value to declare if the ouput is json string for
-                            the json report (True) or html string for html
-                            report (False)
-        :type json_report: bool
-
         :rtype: str
         :returns:  A string that corresponds to HTML that can be embedded in the
-                   html report or the JSON Object that can be embedded into json
-                   report, determine by the json_report parameter.
+                   html report.
         """
         pass
 
@@ -99,6 +93,12 @@ class AnalysisInterface(abc.ABC):
     @abc.abstractmethod
     def get_name():
         """Return name of analysis"""
+        pass
+
+    @staticmethod
+    @abc.abstractmethod
+    def get_json_string_result():
+        """Return json_string_result"""
         pass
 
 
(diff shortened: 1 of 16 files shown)


src/fuzz_introspector/json_report.py
```diff
@@ -0,0 +1,74 @@
...
+                None,
+                proj_profile,
+                profiles,
+                None,
+                coverage_url,
+                None,
+                True
+            )
+
+    report_name = "sink_coverage.json"
```
DavidKorczynski, line 68: This is a good approach in terms of a dedicated module for outputting data. However, could we generalise this?We would like to be able to export all of the data in fuzz introspector into a json file. So, instead of creating a `sink_coverage.json` file we should create a file `report_out.json` and have some logic for writing to this file easily.

The sink plugin should then write to this json file in a localtion e.g. 

```
report {
  plugins {
     ...,
     sink : {
        sink data output.
     }
  }
}
```

All of the other results in fuzz-introspector can then be written to the same `.json` file. Basically, we want a `.json` version of everything (or a large subset of) that is shown in the `.html` files.

DavidKorczynski (changes requested):
An initial request for change for now

src/main.py
```diff
@@ -84,6 +84,14 @@ def get_cmdline_parser() -> argparse.ArgumentParser:
         default="c-cpp",
         help="Language of project"
     )
+    report_parser.add_argument(
+        "--analyses-with-json",
```
DavidKorczynski, line 88: Could we change this to something like `output-json`?

src/fuzz_introspector/json_report.py
arthurscchan: This will be handled in separate PR later.

arthurscchan:
By changing the parameter passing to _--analyses-with-json_ and _--analyses-to-run_, it controls what will be included in the json report and the html report. Analyser name added to --analyses-with-json will be executed and generate output in the json report while name added to --analyses-to-run will be included in the html report. same names can be existed in both options in order to include the execution result in both report format.

src/fuzz_introspector/analysis.py
```diff
@@ -82,9 +83,15 @@ def analysis_func(
                            be shown at the top of the report page.
         :type conclusions: List[html_helpers.HTMLConclusion]
 
+        :param json_report: A bool value to declare if the ouput is json string for
+                            the json report (True) or html string for html
```
DavidKorczynski, line 87: Hmm, am not sure if I like that it's either `json` or `html` because it means in order to collect both we need to run the analysis twice. Could we not make it such that the `.json` content is build up, and also potentially the `html` string which are then accessed through a property on the Python class?

src/main.py
arthurscchan: Done

src/fuzz_introspector/json_report.py
```diff
@@ -0,0 +1,103 @@
...
+            proj_profile,
+            analyses,
+            coverage_url
+        )
+        result_dict['report'][analyses] = json.loads(result_str)
+
+    result_str = json.dumps(result_dict)
+    logger.info("Finish handling sections that need json output")
+    # Write the json string to file
+    report_name = "fuzz-introspector.json"
```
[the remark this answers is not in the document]
arthurscchan: Done

src/fuzz_introspector/analysis.py
arthurscchan: Done

src/fuzz_introspector/json_report.py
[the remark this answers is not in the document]
DavidKorczynski: Almost there :) -- Could you put the variable in https://github.com/ossf/fuzz-introspector/blob/main/src/fuzz_introspector/constants.py together with other constants?

src/fuzz_introspector/json_report.py
```diff
@@ -0,0 +1,94 @@
+# Copyright 2021 Fuzz Introspector Authors
```
DavidKorczynski, line 1: Copyright year

DavidKorczynski (changes requested):

DavidKorczynski (approved):
Great stuff, LGTM!

Am excited adding more data to the json output so we can easier get more macro-level insights on e.g. all OSS-Fuzz projects.

Issue #683, opened 2022-12-13
angus-mail failing to extract relevant data in java frontend analysis


DavidKorczynski:
Migrated from https://github.com/ossf/fuzz-introspector/pull/680#issuecomment-1347330552

The `angus-mail` java project is missing data when analysed by the jvm frontend. For example the calltree for the BASE64EncoderStreamFuzzer we get is:
```
Call tree
fuzzerTestOneInput(com.code_intelligence.jazzer.api.FuzzedDataProvider) BASE64EncoderStreamFuzzer linenumber=-1
  consumeRemainingAsBytes() com.code_intelligence.jazzer.api.FuzzedDataProvider linenumber=29
  write(byte[]) com.sun.mail.util.BASE64EncoderStream linenumber=29
```

This is a call to
https://github.com/eclipse-ee4j/angus-mail/blob/46518aad79259d12a115937894ff8f32b86a1a17/core/src/main/java/com/sun/mail/util/BASE64EncoderStream.java#L141-L144

which should be a call to:
https://github.com/eclipse-ee4j/angus-mail/blob/46518aad79259d12a115937894ff8f32b86a1a17/core/src/main/java/com/sun/mail/util/BASE64EncoderStream.java#L93

So we ideally would have more calls in the callgraph. It seems like there is no analysis of the code in the `angus-core.jar` file but the analysis is limited to the `.jar` files for the fuzzer classes.

DavidKorczynski:
With the changes in https://github.com/ossf/fuzz-introspector/pull/684 the logging output I get from a OSS-Fuzz run is the following:

```
Running introspector frontend on ASCIIUtilityFuzzer.class :: {'ASCIIUtilityFuzzer.jar', 'BASE64EncoderStreamFuzzer.jar', 'angus-core.jar'}
Running command: [java -Xmx6144M -cp /fuzz-introspector/frontends/java/target/ossf.fuzz.introspector.soot-1.0.jar ossf.fuzz.introspector.soot.CallGraphGenerator ASCIIUtilityFuzzer.jar:BASE64EncoderStreamFuzzer.jar:angus-core.jar ASCIIUtilityFuzzer fuzzerTestOneInput jdk.:java.:javax.:sun.:sunw.:com.sun.:com.ibm.:com.apple.:apple.awt.]
[Callgraph] Running callgraph plugin
[...]

PR #688, opened 2022-12-14
JVM implementation bug: Fix package inclusion and exclusion bug

arthurscchan:
Add in all include options to force all library classes to be handled. These library classes includes package prefix which are considered as native code, which may accidentally covers some project with package name prefix same as these native package.

Solve issue #683 


DavidKorczynski (approved):
Fantastic!

PR #691, opened 2022-12-14
analysis: temporary fix around branch blocker calculation

DavidKorczynski:
Fix for https://github.com/google/oss-fuzz/issues/9209#issuecomment-1352241356

@Navidem I first tried to have a bit more tight control, but ended up wrapping the body of the loop in a try-catch -- we can debug specific issues from when we see error logging come up. WDYT?

### DIFF
--- a/src/fuzz_introspector/analysis.py
+++ b/src/fuzz_introspector/analysis.py
@@ -544,49 +544,57 @@ def detect_branch_level_blockers(
 
         # We have some sides taken and some not taken sides => there are blockers.
         for blocked_idx in not_taken_sides:
-            blocked_side = blocked_idx
-            blocked_unique_not_covered_com = (
-                llvm_branch.sides[blocked_idx].unique_not_covered_complexity)
-            blocked_unique_reachable_com = (
-                llvm_branch.sides[blocked_idx].unique_reachable_complexity)
-            blocked_reachable_com = llvm_branch.sides[blocked_idx].reachable_complexity
-            blocked_not_covered_com = llvm_branch.sides[blocked_idx].not_covered_complexity
-            side_line = llvm_branch.sides[blocked_idx].pos
-            side_line_number = side_line.split(':')[1].split(',')[0]
-            blocked_unique_funcs = list(
-                llvm_branch.get_side_unique_reachable_funcnames(blocked_idx))
+            try:
+                blocked_side = blocked_idx
+                blocked_unique_not_covered_com = (
+                    llvm_branch.sides[blocked_idx].unique_not_covered_complexity)
+                blocked_unique_reachable_com = (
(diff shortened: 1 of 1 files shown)


src/fuzz_introspector/analysis.py
```diff
@@ -544,49 +544,58 @@ def detect_branch_level_blockers(
...
+                new_blk = FuzzBranchBlocker(blocked_side, blocked_unique_not_covered_com,
+                                            blocked_unique_reachable_com, blocked_unique_funcs,
+                                            blocked_not_covered_com, blocked_reachable_com,
+                                            hitcount_diff, source_file_path, line_number,
+                                            side_line_number, function_name, link)
+                fuzz_blockers.append(new_blk)
+            except Exception:
+                # Temporary fix for https://github.com/google/oss-fuzz/issues/9209
+                # This should be refined and made more precise.
+                logger.debug("Branch-blocker: exception occurred in iteration of not-taken-sides")
```
Navidem, line 597: Let's log the `llvm_branch.sides` as well to help finding the cause.

Navidem (approved):

Navidem:
> Fix for [google/oss-fuzz#9209 (comment)](https://github.com/google/oss-fuzz/issues/9209#issuecomment-1352241356)
> 
> @Navidem I first tried to have a bit more tight control, but ended up wrapping the body of the loop in a try-catch -- we can debug specific issues from when we see error logging come up. WDYT?
> 
> Signed-off-by: David Korczynski [david@adalogics.com](mailto:david@adalogics.com)

I was expecting trail_build to catch such cases if it was a common one, or maybe `nDPI ` is a little different. Have to check it.

DavidKorczynski:
> 



> > Fix for [google/oss-fuzz#9209 (comment)](https://github.com/google/oss-fuzz/issues/9209#issuecomment-1352241356)
> > @Navidem I first tried to have a bit more tight control, but ended up wrapping the body of the loop in a try-catch -- we can debug specific issues from when we see error logging come up. WDYT?
> > Signed-off-by: David Korczynski [david@adalogics.com](mailto:david@adalogics.com)
> 
> I was expecting trail_build to catch such cases if it was a common one, or maybe `nDPI ` is a little different. Have to check it.

It did not catch it because of the `|| true` part when calling introspector. See details in the second half here: https://github.com/google/oss-fuzz/issues/9209#issuecomment-1352241356

src/fuzz_introspector/analysis.py
DavidKorczynski: Done, please check

Navidem: Thanks!

Navidem:
Let's merge #695, to prevent the failure and log the culprit.

DavidKorczynski:
> Let's merge #695, to prevent the failure and log the culprit.

done

PR #692, opened 2022-12-14
Fix class/instance implementation

arthurscchan:
Better approach for #690 


DavidKorczynski:
let's just hold on to this one for a day: I would like to first get this through https://github.com/ossf/fuzz-introspector/pull/691 so we can  merge: https://github.com/google/oss-fuzz/pull/9208 in order fix this https://github.com/google/oss-fuzz/issues/9209 and then

arthurscchan:
This is outdated-implementation, closing it.

PR #695, opened 2022-12-15
Add sanity checks to capture potential out-of bounds index

Navidem:
Addressing https://github.com/google/oss-fuzz/issues/9209#issuecomment-1352241356

A replacement for temporary solution in #691

Signed-off-by: Navid Emamdoost <navid.emamdoost@gmail.com>

### DIFF
--- a/src/fuzz_introspector/analysis.py
+++ b/src/fuzz_introspector/analysis.py
@@ -542,62 +542,61 @@ def detect_branch_level_blockers(
         if len(taken_sides) == 0 or len(not_taken_sides) == 0:
             continue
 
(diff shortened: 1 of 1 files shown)


Navidem:
The problem in https://github.com/google/oss-fuzz/issues/9209#issuecomment-1352241356 is that in some cases, the branch profiles loaded from coverage reports is not consistent with LLVM branch profile. More specifically in number of branch sides. Definitely this is a bug in coverage parsing when identifying branch sides based.

DavidKorczynski (approved):

PR #699, opened 2022-12-15
Json report: Make json string result an instance variable

arthurscchan:
This PR aims to revamp the json string result object and make it an instance variable. Then fix the logic to store those analyser instance that request to be included in the resulting json report.


DavidKorczynski (approved):
Thanks!

Issue #693, opened 2022-12-14
Find out why CI didn't catch recent regressions


DavidKorczynski:
- https://github.com/ossf/fuzz-introspector/pull/686 introduced a bug that was fixed here https://github.com/ossf/fuzz-introspector/pull/690

- Another regression was recently introduced and I think here: https://github.com/ossf/fuzz-introspector/pull/657 Details of the issue is here https://github.com/google/oss-fuzz/issues/9209#issuecomment-1352241356

I feel these issues should both have been caught by the CI, either the fuzzing or the end-to-end tests we have. We should improve the CI to detect such issues being introduced.

PR #701, opened 2022-12-15
fuzz: specify arg values in fuzz_report_generation

DavidKorczynski:
This is one of the reason some issues have not been reported by the fuzzer. https://github.com/ossf/fuzz-introspector/pull/686 introduced a new argument to `run_analysis_on_dir` which caused an argument in the fuzzer (the last argument, `False`) to reference a new argument. The effect is that the underlying loading of yaml profiles were multi-threaded, which consequntly do not propagate exceptions.

This commit changes it to specify arguments when calling `run_analysis_on_dir` so future scenarios won't switch the arguments as such. Note, I don't think this would have detected the branch blocker issues as this happened before the above change.

Ref: https://github.com/ossf/fuzz-introspector/issues/693

### DIFF
--- a/src/fuzz_introspector/code_coverage.py
+++ b/src/fuzz_introspector/code_coverage.py
@@ -26,6 +26,7 @@ from typing import (
 )
 
 from fuzz_introspector import utils
+from fuzz_introspector import exceptions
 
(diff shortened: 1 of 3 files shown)


Navidem (approved):
LGTM

PR #705, opened 2022-12-15
workflows: speed up jvm-test by limiting test cases

DavidKorczynski:


.github/workflows/jvm-test.yml
```diff
@@ -15,4 +15,4 @@ jobs:
         java-version: '11'
     - name: JVM-Test
       run: |
-        cd tests/java && ./buildAll.sh && ./runTest.sh
+        cd tests/java && ./buildAll.sh && ./runTest.sh test1 && ./runTest.sh test2
```
arthurscchan, line 18: Actually, this was some old code checking if the compile and general execution is running without exception and did not analyse the result. They are now covered by the unit testing with more details matching of the result. Not sure if this is still needed.

arthurscchan (approved):

.github/workflows/jvm-test.yml
arthurscchan: The only reason to keep this CI is because it is run in Java based environment and are good to open for additional "Java only" unit testing to this CI.

DavidKorczynski: Ah, just saw your second comment after having just removed it. Let's just remove it now to simplify things -- if we need it down the line for additional java only unit testing then we'll do that. I think it's good to keep it in the main unit testing workflow for now, also because that workflow is closer to how the code is run in OSS-Fuzz

AdamKorcz (approved):

PR #707, opened 2022-12-16
test: fuzz: use hack to make CIFuzz able to reproduce

DavidKorczynski:
Avoid instrumenting in reproducer runs.


DavidKorczynski:
The ClusterFuzzLite job is now reporting the issue correctly and is overcoming the issue here https://github.com/google/oss-fuzz/issues/9222 
I think what really did the difference is the `report-unreproducible-crashes: true`.

Will fix the issues reported in the code in another PR.

AdamKorcz (approved):

PR #709, opened 2022-12-16
JVM bug: Fix naming problem

arthurscchan:
Fix naming output problem of Soot code for the entry method. Also changing the fuzz-introspector code to match the updated entry function names. Also include the entry function to the reached function list.

Last but not least, update the unit testing to match the changed name.

Relies on #702 


DavidKorczynski (approved):
LGTM

The error reported by ClusterFuzzLite CI job is a false positive and not related to this PR

PR #710, opened 2022-12-16
JVM patch: Fix missing coverage for entry method

arthurscchan:
Change handling of entry point method name and add the entrypoint method to the reached method list.

This PR relies on #709 because it only work on the new version of the data and data.yaml generation and may need to fix some of the unit testing before merging.


arthurscchan:
Merge the PR to #709, cause the change are inter-affect by each other.

PR #711, opened 2022-12-16
data_loader: wrap data loading to catch unicode errors

DavidKorczynski:
This should fix a bug found by the report fuzzer.


DavidKorczynski:
Am not why clusterfuzz-lite reports finding a crash: there's no crash, stack trace and the crashing input is empty. This is good to go.

AdamKorcz (approved):

PR #735, opened 2022-12-29
Sink Coverage Fix naming issue

arthurscchan:
Fix sink and all other function naming

### DIFF
--- a/src/fuzz_introspector/analyses/sinks_analyser.py
+++ b/src/fuzz_introspector/analyses/sinks_analyser.py
@@ -205,9 +205,8 @@ class Analysis(analysis.AnalysisInterface):
         callsite: cfg_load.CalltreeCallsite
     ) -> str:
         """
-        Handling function name and package name combo
-        and avoid missing or double package name existed in
-        the final function name comparison.
+        Add package name to uniquly identify functions
+        in different package.
         """
         func_name = f"{callsite.dst_function_name}"
         if func_name.startswith("["):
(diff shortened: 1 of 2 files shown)


src/fuzz_introspector/utils.py
```diff
@@ -140,7 +140,11 @@ def demangle_cpp_func(funcname: str) -> str:
 
 
 def demangle_jvm_func(package: str, funcname: str) -> str:
-    return "[%s].%s" % (package, funcname)
+    """Add package class name to uniquly identify jvm functons"""
+    if funcname.startswith("["):
```
DavidKorczynski, line 144: Does function name ever start with "["? Do you have an example?

arthurscchan: This is to avoid double assignment of package name. Make it safe when some already processed name will not be processed again.

src/fuzz_introspector/analyses/sinks_analyser.py
```diff
@@ -198,15 +200,30 @@ def _retrieve_data_list(
 
         return (callsite_list, function_list)
 
+    def _handle_function_name(
+        self,
+        callsite: cfg_load.CalltreeCallsite
+    ) -> str:
+        """
+        Handling function name and package name combo
```
DavidKorczynski, line 208: This function description is a bit unclear (what does "handling" mean?), and has some typos -- could you rewrite it?

DavidKorczynski (changes requested):

src/fuzz_introspector/analyses/sinks_analyser.py
arthurscchan: yes

DavidKorczynski (approved):

PR #739, opened 2022-12-29
Rename analyses to avoid Analysis class

DavidKorczynski:
All analyses have an `Analysis` class, which is slightly annoying for python documentation (https://fuzz-introspector.readthedocs.io/) as it shows all of them with the same name. This changes the general names to plugin-specific names, and then has a list in analyses/__init__.py with all of the various plugin names. This improves documentation and I think also makes the code better.


DavidKorczynski:
ClusterFuzzLite issue is a false positive

AdamKorcz (approved):

PR #748, opened 2023-01-04
refactor json reporting

DavidKorczynski:
Remove the use of write_to_summary in the utils.py module, and make more convenient wrapper functions in json_report.py. This also overcomes some existing limitations where we basically overwrote various key/value pairs in the report.


DavidKorczynski:
ClusterFuzzLite is a false positive

AdamKorcz (approved):

PR #750, opened 2023-01-04
Output all functions table to json report.

DavidKorczynski:
This is needed for things such as comparing fuzz introspector runs and for creating scripts that gather overviews of a lot of fuzz introspector reports.


Navidem (approved):
LGTM

PR #757, opened 2023-01-10
html_report: fix value writing to json report

DavidKorczynski:
The existing `row` is in HTML format whereas we want it raw.


Navidem (approved):
LGTM

PR #762, opened 2023-01-12
html report: show calltree indices in fuzz blocker table

DavidKorczynski:
The indices are not shown anywhere in the table, but they are quite useful for correlating the blocker with the bitmap overview.


Navidem (approved):
LGTM

PR #764, opened 2023-01-16
Sink Analyser: Showing callpath for sink functions

arthurscchan:
Related to issue #670 

### DIFF
--- a/src/fuzz_introspector/datatypes/function_profile.py
+++ b/src/fuzz_introspector/datatypes/function_profile.py
@@ -45,11 +45,16 @@ class FunctionProfile:
         self.edge_count = elem['EdgeCount']
         self.cyclomatic_complexity = elem['CyclomaticComplexity']
         self.functions_reached = utils.load_func_names(elem['functionsReached'], False)
-        self.functions_called = utils.load_func_names(elem['functionsReached'], False)
         self.function_uses = elem['functionUses']
         self.function_depth = elem['functionDepth']
         self.constants_touched = elem['constantsTouched']
         self.branch_profiles = self.load_func_branch_profiles(elem['BranchProfiles'])
+
+        # Duplication of functions_reached to keep the original set of call tree
+        # relationship for further processing and analysis. This could avoid lost
+        # of callpath tree information when functions_Reached is further propogated
+        # by later operations.
+        self.functions_called = utils.load_func_names(elem['functionsReached'], False)
 
         # Temporary handle for unreadable library method (JVM)
         # (jar missing or purposely ignored)
--- a/src/fuzz_introspector/datatypes/function_profile.py
+++ b/src/fuzz_introspector/datatypes/function_profile.py
@@ -45,6 +45,7 @@ class FunctionProfile:
         self.edge_count = elem['EdgeCount']
         self.cyclomatic_complexity = elem['CyclomaticComplexity']
         self.functions_reached = utils.load_func_names(elem['functionsReached'], False)
+        self.functions_called = utils.load_func_names(elem['functionsReached'], False)
         self.function_uses = elem['functionUses']
         self.function_depth = elem['functionDepth']
         self.constants_touched = elem['constantsTouched']
--- a/src/fuzz_introspector/datatypes/function_profile.py
+++ b/src/fuzz_introspector/datatypes/function_profile.py
@@ -50,10 +50,10 @@ class FunctionProfile:
         self.constants_touched = elem['constantsTouched']
         self.branch_profiles = self.load_func_branch_profiles(elem['BranchProfiles'])
 
-        # Duplication of functions_reached to keep the original set of call tree
-        # relationship for further processing and analysis. This could avoid lost
-        # of callpath tree information when functions_Reached is further propogated
-        # by later operations.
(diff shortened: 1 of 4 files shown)


src/fuzz_introspector/datatypes/project_profile.py
```diff
@@ -260,6 +260,29 @@ def get_complexity_summaries(self) -> Tuple[int, int, int, float, float]:
             unreached_complexity_percentage
         )
 
+    def get_function_callpaths(
+        self,
+        target_function: function_profile.FunctionProfile
+    ) -> List[List[function_profile.FunctionProfile]]:
+        """
+        Recrusively resolve the incoming reference of a function
```
DavidKorczynski, line 268: nit: Recursively

DavidKorczynski (changes requested):
Could you add some comments in the PR showing an example? Could we add a unit test as well?

src/fuzz_introspector/datatypes/project_profile.py
arthurscchan: Fixed

arthurscchan:
The callpath is discovered by reverse recursive searching through the incoming_reference parameter in the target function profile. It loops through each of the incoming_reference of a target function to their parents functions, and so on, until reaching a top most functions which has no incoming_reference. That function with no incoming reference indicates the outtest functions of the project that can reach the target function, allowing a fuzzer to build and specifically target the target function.

arthurscchan:
The unit testing of this PR will be added in separate PR. This PR aims to provide additional callpath information for a specific target functions, that is those sink functions discovered by the sink analyser.

arthurscchan:
Example:
If there is a sink function with the following callpath returns
[['A'->'B'->'C'], ['D'->'E'->'F'], ['G'->'E'->'F']]
It means
Function C and F directly calls the sink functions and the call path for function C in the project is A->B->C and there are two call paths for function F through function D or function G, passing through function E.

The recursive logic digs up the incoming references of the sink function, which are A B C D E F G in this case. All of the functions in the "incoming references list" has at least 1 call path reaching the target sink function. 

The logic then do a reverse search to search through the calltree in order to reach function A D G which does not have further incoming_reference, which are assuming to be the outtermost level of functions in the project which could be used to build fuzzers to fuzz the target sink function

src/fuzz_introspector/datatypes/function_profile.py
```diff
@@ -45,6 +45,7 @@ def __init__(self, elem: Dict[Any, Any]) -> None:
         self.edge_count = elem['EdgeCount']
         self.cyclomatic_complexity = elem['CyclomaticComplexity']
         self.functions_reached = utils.load_func_names(elem['functionsReached'], False)
+        self.functions_called = utils.load_func_names(elem['functionsReached'], False)
```
DavidKorczynski, line 48: This looks to be the same as `functions_reached`? Any reason not to use that?

DavidKorczynski (changes requested):

src/fuzz_introspector/datatypes/function_profile.py
arthurscchan: Yes, I duplicate it on-purpose. Because self.functions_reached got further processed in 
https://github.com/ossf/fuzz-introspector/blob/main/src/fuzz_introspector/datatypes/fuzzer_profile.py#L231
for further analysis and the original set of "functionsReached" is lost in progress which is needed to build up the callpath tree for the sink analyser.

DavidKorczynski: Can you leave a comment in the code saying this?

arthurscchan: OK

arthurscchan:
Sample for textual path to sink functions. Considering graphical or colourful display of the tracing path in later development.
test

src/fuzz_introspector/datatypes/function_profile.py
```diff
@@ -50,6 +50,12 @@ def __init__(self, elem: Dict[Any, Any]) -> None:
         self.constants_touched = elem['constantsTouched']
         self.branch_profiles = self.load_func_branch_profiles(elem['BranchProfiles'])
 
+        # Duplication of functions_reached to keep the original set of call tree
+        # relationship for further processing and analysis. This could avoid lost
+        # of callpath tree information when functions_Reached is further propogated
+        # by later operations.
```
DavidKorczynski, line 56: Could you go over this text again? There are a few typos (propogated --> propagated, "functions_Reached" --> functions_reached

DavidKorczynski (changes requested):

DavidKorczynski:
> Sample for textual path to sink functions. Considering graphical or colourful display of the tracing path in later development. test

Once the typos are fixed then let's merge this in. But we should improve the graphics: in the above picture it's already fairly difficult to digest the information and I think the size of the information will explode with other targets

src/fuzz_introspector/datatypes/function_profile.py
arthurscchan: Done

DavidKorczynski (approved):

PR #767, opened 2023-01-17
Sink Analyser: Extract sink function to config file

arthurscchan:


DavidKorczynski (changes requested):
What's the reason for this? I'd probably prefer to have it in the code rather than a config file, unless something is blocking it for being in the code?

arthurscchan:
Not needed anymore.

PR #768, opened 2023-01-17
Sink Analyser: Fuzzer hit information and function blocker

arthurscchan:
Add features to show fuzzer hitting information for sink functions. Also provide callpath suggestion for creating fuzzers for specific sink function.


src/fuzz_introspector/analyses/sinks_analyser.py
```diff
@@ -288,6 +288,27 @@ def _filter_function_list(
...
 
+    def _retrieve_fuzzer_hitcount(
+        self,
+        functions function_profile.FunctionProfile,
+        coverage: code_coverage.CoverageProfile
+    ) -> int:
+        """
+        Analyse the project coverage and calculate the hit
+        count for target function. This information also shows
+        if the target function is covered by a specific fuzzers
```
DavidKorczynski, line 299: typo: `fuzzer` minus `s`

DavidKorczynski (changes requested):

src/fuzz_introspector/analyses/sinks_analyser.py
arthurscchan: fixed

DavidKorczynski:
Could you provide a screenshot?

arthurscchan:
Currently, if the sink function has no fuzzer reached in runtime, it will dig up all possible callpaths and determine which function is the blocker, that is the function that fail to call down the callpath, together with its information for reference.

DavidKorczynski (approved):
LGTM

We should add some constraints to avoid showing the HTML as of now since they are not fully ready for displaying on e.g. oss-fuzz. Will do this in a follow-up pr.

ClusterFuzzLite pr is the false positive we've been experiencing lately. No real issue.

PR #772, opened 2023-01-18
Sink Analyser: Pretty print callpath result

arthurscchan:


arthurscchan:
Currently, the callpath for each parent function of the sink function will be sorted and only top 20 of the callpath link is displayed. The link will direct the user to another new webpage showing the callpath and some related information.
The sorting is based on the length of the callpath, the 20 shallowest callpath will be displayed, indicate possible API calls that are closest to the target sink function.

DavidKorczynski (approved):

DavidKorczynski:
Awesome! Moving the sink details page into individual pages is nice and I think we can use this for additional features in the future.

PR #784, opened 2023-01-27
test CI

DavidKorczynski:
Don't merge. Testing if we still run into a false positive in ClusterFuzzLite: https://github.com/google/oss-fuzz/issues/9318#issuecomment-1406929101

Signed-off-by: DavidKorczynski <david@adalogics.com>


DavidKorczynski:
all looks to be working!

PR #790, opened 2023-01-31
Sink Analyser: Save sink analyser testing data generation script

arthurscchan:
Issue #670 


DavidKorczynski (changes requested):
Awesome! One minor request: could you move this into `tools/sink-analyser` instead of `test/sink-analyser`? I'd like for us to gather all tools of some kind in a separate directory.

tools/sink-analyser/retrieve_project_link.py
```diff
@@ -0,0 +1,31 @@
...
+
+import requests
+import lxml.html
+import json
+
+data = requests.get('https://oss-fuzz-introspector.storage.googleapis.com/index.html').text
+html = lxml.html.fromstring(data)
+
+link_map = dict()
+for li in html.getchildren()[1].getchildren()[2]:
```
DavidKorczynski, line 24: The HTML changed recently and looks different now: https://oss-fuzz-introspector.storage.googleapis.com/index.html -- I assume the HTML  change may have broken the script?

Am not sure what the smartest thing is here in order to make it more general. However, for now, could you update the script so it works with the latest?

DavidKorczynski (changes requested):

tools/sink-analyser/retrieve_project_link.py
arthurscchan: changed

DavidKorczynski (approved):

PR #791, opened 2023-02-01
JVM Implementation: Fix soot logic to include callsite for all processed methods

arthurscchan:
Fix soot logic to loop through all statement in processed methods to record its method invocation. This can help to provide callsite information and line number of every methods that has a parent function (with source code provided).

### DIFF
--- a/frontends/java/src/main/java/ossf/fuzz/introspector/soot/CallGraphGenerator.java
+++ b/frontends/java/src/main/java/ossf/fuzz/introspector/soot/CallGraphGenerator.java
@@ -33,6 +33,7 @@ import java.util.Set;
 import org.apache.commons.lang3.StringUtils;
 import ossf.fuzz.introspector.soot.yaml.BranchProfile;
 import ossf.fuzz.introspector.soot.yaml.BranchSide;
+import ossf.fuzz.introspector.soot.yaml.Callsite;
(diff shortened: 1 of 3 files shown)


frontends/java/src/main/java/ossf/fuzz/introspector/soot/CallGraphGenerator.java
```diff
@@ -674,6 +647,67 @@ private String mergeClassName(Set<String> classNameSet) {
     return mergedClassName.toString();
   }
 
+  private Callsite handleStatement(Stmt stmt, String sourceFilePath) {
```
DavidKorczynski, line 650: Could you make the function names more descriptive and some descriptive comments/function-level doc? `handle` is too generic since it does not say much about the semantics of the function.

DavidKorczynski (changes requested):

frontends/java/src/main/java/ossf/fuzz/introspector/soot/CallGraphGenerator.java
arthurscchan: Changed

DavidKorczynski (approved):
LGTM

Please run docs through a spellcheck in the future. Will handle the typos here after the merge.

PR #802, opened 2023-02-03
JVM Integration: Fix package name discovery for project with fuzzer using package

arthurscchan:
Some project in oss-fuzz have java fuzzers not using default package. This PR aims to fix the patching script in order to handle those cases and correctly discover the needed full class name with package.
The package name are discovered by examine the class file path in the jar file. All class files in a jar file will be stored under subdirectories reassemble the class package name. For example a class with package A.B.C.D are stored in A/B/C/D directory in the jar file.


arthurscchan:
For example, the rhino project in https://github.com/google/oss-fuzz/tree/master/projects/rhino which have fuzzers in the package ossfuzz instead of default package.

frontends/java/oss-fuzz-main.py
```diff
@@ -56,25 +57,46 @@ def find_fuzz_targets(path):
...
+
     # Create relevant .jar file if it is not there.
-    jar_file = classfile.replace(".class", ".jar")
     if not os.path.exists(jar_file):
       subprocess.check_call("jar cvf %s %s" % (jar_file, classfile), shell=True)
 
+    # Extract jar file for package name discovery.
+    temp_dir = os.path.join(path, 'temp-jar')
+    os.mkdir(temp_dir)
+    subprocess.check_call("cd %s && jar xvf %s/%s" % (temp_dir, path, jar_file), shell=True)
```
DavidKorczynski, line 79: Please use `cwd` from https://docs.python.org/3/library/subprocess.html#subprocess.check_call instead of `cd && ..`

DavidKorczynski (changes requested):

frontends/java/oss-fuzz-main.py
arthurscchan: done

frontends/java/oss-fuzz-main.py
```diff
@@ -56,25 +57,46 @@ def find_fuzz_targets(path):
...
+
     # Create relevant .jar file if it is not there.
-    jar_file = classfile.replace(".class", ".jar")
     if not os.path.exists(jar_file):
       subprocess.check_call("jar cvf %s %s" % (jar_file, classfile), shell=True)
 
+    # Extract jar file for package name discovery.
+    temp_dir = os.path.join(path, 'temp-jar')
+    os.mkdir(temp_dir)
+    subprocess.check_call("jar xvf %s/%s" % (path, jar_file), shell=True, cwd=temp-dir)
```
DavidKorczynski, line 79: Do we have to use a command to untar a file? Can we use https://docs.python.org/3/library/tarfile.html instead?

arthurscchan: Fixed the logic, now it is not necessary to examine the jar file.

DavidKorczynski (approved):

PR #809, opened 2023-02-06
JVM Integration: Fix bug for frontend java code

arthurscchan:
Fix some bug for JAVA frontend code, also fix some logic to decrease the size of the output data and yaml file.


DavidKorczynski (approved):
LGTM

For future references, could you please add some more descriptive comments with the PR? For  instance, instead of "Fix bug" it would be good to have a description that: (1) explains the substance of the bug and (2) gives an example where this bug occurs

Issue #817, opened 2023-02-10
add java installation instructions to readthedocs


DavidKorczynski:
Complete this section https://fuzz-introspector.readthedocs.io/en/latest/installation-instructions/installation.html#java

DavidKorczynski:
Is this solved @arthurscchan ?

arthurscchan:
Solved in #824

PR #824, opened 2023-02-13
JVM integration: Add Java section for Fuzz Introspector installing documentation

arthurscchan:
Fill in the section for Java language in the Fuzz Introspector installing documentation.
For issue #817 


DavidKorczynski (changes requested):
Could we add an example that shows the details of the frontend? At the moment it's hidden within `runTests.sh` where it would be nice to see the examples of the verbose commands needed.

DavidKorczynski (approved):

PR #853, opened 2023-02-23
Auto-Fuzz: Fix concrete subclass searching logic

arthurscchan:
For Java program, it is possible to specify a abstract superclass or interface as the argument type and all object of its subclass or class implemnting that interface are all valid object for that argument type. There is possibility that there are more than one possible options for the concrete object, this PR aims to generate object creation statement for each of the possibility in order to cover more fuzzer choice and targets. This PR also adds in a new constant for limiting the number of target generated for each project to increase efficiency.


DavidKorczynski:
Please resolve conflicts following merge of https://github.com/ossf/fuzz-introspector/pull/852

DavidKorczynski (approved):

PR #856, opened 2023-02-24
Auto-fuzz: Add heuristic3 for java

arthurscchan:
The new heuristic 3 considers some cases where constructor call is not possible. In some of the classes, they provide some methods, which we normally named them as factory methods, to generate an object instance of a specific class. In some of these cases, the constructor are protected and developers are required to use those factory methods to create an object instance which may contain hidden settings or singleton control. These factory methods can be static or non-static. This heuristic focuses on static factory methods and calls them to retrieve an object instance before invoking an instance method. The focus for non-static factory methods are processed in heuristic 4.


DavidKorczynski:
Could you please rebase?

DavidKorczynski (approved):

PR #860, opened 2023-02-28
Auto-fuzz: Add tqdm module to requirement for python dependency

arthurscchan:
The post processing of auto-fuzz uses the tqdm python module to display loading bar when the process check the auto generated fuzzer. This PR adds the tqdm module to the python dependency requirement list to enable it.


DavidKorczynski (approved):
LGTM -- thanks! Down the line maybe we should add a `requirements.txt` for `auto-fuzz` itself, in the event a lot of dependencies will be for `auto-fuzz` only.

PR #862, opened 2023-02-28
Auto-fuzz: Fix java autofuzz build fail

arthurscchan:
The original code assumes that the target project uses maven structure, but there exists quite a number of projects that actually adopt Gradle project structure instead. This PR add logic to the auto-fuzz manager and base file builder in order to accommodate gradle projects. The new code will check if the project contains pom.xml or build.gradle which are the build property files for maven and gradle project respectively and build with the corresponding builder.

### DIFF
--- a/tools/auto-fuzz/manager.py
+++ b/tools/auto-fuzz/manager.py
@@ -206,7 +206,8 @@ def _maven_build_project(basedir, projectdir):
 
     # Set environment variable
     env_var = os.environ.copy()
-    env_var['PATH'] = os.path.join(basedir, "apache-maven-3.6.3", "bin") + ":" + env_var['PATH']
+    env_var['PATH'] = os.path.join(basedir, "apache-maven-3.6.3",
+                                   "bin") + ":" + env_var['PATH']
 
     # Build project with maven
     cmd = [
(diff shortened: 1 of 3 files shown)


tools/auto-fuzz/manager.py
```diff
@@ -196,13 +198,71 @@ def run_static_analysis_python(git_repo, basedir):
...
+    ]
+    try:
+        subprocess.check_call(" && ".join(cmd),
+                              shell=True,
+                              timeout=1800,
+                              stdout=subprocess.DEVNULL,
+                              stderr=subprocess.DEVNULL,
+                              env=env_var,
+                              cwd=projectdir)
+    except subprocess.TimeoutExpired:
```
DavidKorczynski, line 254: I've had troubles where this breaks build because this command fails. We would like to avoid some commands failing causing a large-scale experiment breaking.

tools/auto-fuzz/manager.py
```diff
@@ -503,15 +571,23 @@ def autofuzz_project_from_github(github_url,
                          oss_fuzz_base_project.project_name)):
         return False
 
-    # If this is a jvm target download maven once so we don't have to do it
-    # for each proejct.
+    # If this is a jvm target download maven and gradle once so we don't
+    # have to do it for each proejct.
     if language == "jvm":
+        # Download Maven
         MAVEN_URL = "https://downloads.apache.org/maven/maven-3/3.6.3/binaries/apache-maven-3.6.3-bin.zip"
```
DavidKorczynski, line 578: Also these types of constants, please keep them in the `constants.py` module or elsewhere e.g. at the top of the module

PR #869, opened 2023-03-01
Auto-fuzz: Move maven and gradle properties to constants

arthurscchan:
Move the maven and gradle properties to constants.py, this could make it easier to udpate the path and download links.


DavidKorczynski (approved):
Thanks!

PR #870, opened 2023-03-01
Auto-fuzz: Fix recursion

arthurscchan:
In the Java auto fuzzing, there is a temporary measure added before to limit the number of recursion. When the maximum number of recursion is reached, the searching of possible constructor will stop and return an empty constructor as a fail safe option. This setting could eliminate the maximum recursion count reached problem, but it also makes some of the fuzzers fail to use the fuzzing data because empty constructor are used. 
With the fix in #862 It eliminates the error on repeat process of the same method. This will limit the number of recursion and the needs of the recursion count, thus this PR aims to remove the recursion count feature to allow a better fuzzer generation process with random fuzz data.


DavidKorczynski (approved):
Great, thanks!

PR #877, opened 2023-03-02
Auto-fuzz: Heuristic8 for java

arthurscchan:
Following the change in PR #875, this PR create heuristic 8 to handle non-static method calls which use enum object as part of its arguments.

### DIFF
--- a/tools/auto-fuzz/fuzz_driver_generation_jvm.py
+++ b/tools/auto-fuzz/fuzz_driver_generation_jvm.py
@@ -960,10 +960,10 @@ def generate_possible_targets(proj_folder, max_target):
         yaml_dict = yaml.safe_load(stream)
 
     possible_targets = []
-    #    _generate_heuristic_1(yaml_dict, possible_targets, max_target)
-    #    _generate_heuristic_2(yaml_dict, possible_targets, max_target)
-    #    _generate_heuristic_3(yaml_dict, possible_targets, max_target)
-    #    _generate_heuristic_4(yaml_dict, possible_targets, max_target)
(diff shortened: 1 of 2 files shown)


tools/auto-fuzz/fuzz_driver_generation_jvm.py
```diff
@@ -809,9 +960,10 @@ def generate_possible_targets(proj_folder, max_target):
...
 
     possible_targets = []
-    _generate_heuristic_1(yaml_dict, possible_targets, max_target)
-    _generate_heuristic_2(yaml_dict, possible_targets, max_target)
-    _generate_heuristic_3(yaml_dict, possible_targets, max_target)
-    _generate_heuristic_4(yaml_dict, possible_targets, max_target)
+    #    _generate_heuristic_1(yaml_dict, possible_targets, max_target)
+    #    _generate_heuristic_2(yaml_dict, possible_targets, max_target)
+    #    _generate_heuristic_3(yaml_dict, possible_targets, max_target)
+    #    _generate_heuristic_4(yaml_dict, possible_targets, max_target)
```
DavidKorczynski, line 966: I assume we want to keep these?

DavidKorczynski (changes requested):

tools/auto-fuzz/fuzz_driver_generation_jvm.py
DavidKorczynski: i.e. avoid commenting out?

arthurscchan: Changed.

DavidKorczynski (approved):

PR #878, opened 2023-03-03
Auto-fuzz: Fix check error

arthurscchan:
Some gradle project build will fail for spotless java check target. The main reason of that is the use of deprecated API which are legal java statement but not allowed by the check. This PR excludes the spotless java check and javadoc check to avoid the error and to increase speed from skipping the check. 
Also, a minor fix in manager.py to give execution permission to gradle wrapper is included in this PR.


arthurscchan:
Wrong implementation

PR #898, opened 2023-03-08
Auto-fuzz: Fix primitive object array

arthurscchan:
In java, primitive types and their class objects will auto-box and auto-unbox, thus they can be used interchangeably. But an array of primitive type and array of its class object cannot be auto-boxed or auto-unboxed because that mechanism is only possible for each single primitive type. The current argument handling logic assumes those arrays are interchangeably and by default jazzer are only returning arrays of primitive types. This will cause compile error if class object array type are used. This PR fixes this error by using Apache Common Lang3 library to translate primitive type arrays to their respective class object arrays to avoid the compilation error

### DIFF
--- a/tools/auto-fuzz/base_files.py
+++ b/tools/auto-fuzz/base_files.py
@@ -158,8 +158,11 @@ do
   RUNTIME_CLASSPATH=$RUNTIME_CLASSPATH\$this_dir/$(basename $JARFILE):
 done
 
-BUILD_CLASSPATH=$BUILD_CLASSPATH:$JAZZER_API_PATH
-RUNTIME_CLASSPATH=$RUNTIME_CLASSPATH:\$this_dir
+# Retrieve apache-common-lang3 library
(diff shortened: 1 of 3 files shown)


tools/auto-fuzz/base_files.py
```diff
@@ -158,8 +158,11 @@ def gen_builder_1_jvm():
   RUNTIME_CLASSPATH=$RUNTIME_CLASSPATH\$this_dir/$(basename $JARFILE):
 done
 
-BUILD_CLASSPATH=$BUILD_CLASSPATH:$JAZZER_API_PATH
-RUNTIME_CLASSPATH=$RUNTIME_CLASSPATH:\$this_dir
+# Retrieve apache-common-lang3 library
```
DavidKorczynski, line 161: Could we add some more documentation what it's used for? Basically a refined version of the PR text but in the code itself.

DavidKorczynski (changes requested):

tools/auto-fuzz/base_files.py
arthurscchan: Done.

DavidKorczynski (approved):

Issue #794, opened 2023-02-01
Discrepancy between reachability and coverage for Python projects


DavidKorczynski:
https://oss-fuzz-introspector.storage.googleapis.com/index.html recently got a facelift, and this made some issues obvious. One is that code coverage is often a lot higher than reachability for Python projects. We should investigate why this is the reason and come up with a solution.

DavidKorczynski:
One of the main reasons for this discrepancy is that the reachability analysis is much more focusesd in comparison to the code coverage analysis. For example, the following fuzzer:

```
import mod1
import mod2
import mod3

import atheris

def TestOneInput(data):
  return

# Code to trigger atheris
# ...
```
Will have a lot of code coverage in the modules `mod1`, `mod2` and `mod3` since the coverage collection will happen *before* the modules are imported, and the import statements will cause a lot of code execution in the modules (such as registrering each `def` in the module) where this is not considered in the reachability analysis, as the reachability only considers that code within the `TestOneInput`, which in this case will be 0.

A couple of things that can be done:
- enable code coverage to be started within the fuzzer entrypoint. The negative side of this is that we will not achieve full code coverage unless `imports` are specified there etc.
- make some type of middle-ground where it's clear what code coverage is achieved by the fuzzer versus the e.g. code that happens *before* the fuzzer executes.
- Include reachability analysis based on the *full* fuzzer file, versus only the fuzzer entrypoint.

PR #932, opened 2023-03-23
datatypes: project-profile: refine function source list

DavidKorczynski:


DavidKorczynski:
This is kinda related to https://github.com/ossf/fuzz-introspector/issues/794 -- in this case we will exclude more functions in the stats table, but in particular the number of functions with code coverage. This will help a lot for Python projects, where currently builtin's are included in the list as well as functions from third-party dependencies that not really part of a given package's source.

AdamKorcz (approved):

PR #937, opened 2023-03-27
JVM Integration: Fix logic to allow specifying target package prefix

arthurscchan:
In the current logic, the fuzz-introspector will gather all jar files built and analyse all classes in them and exclude some well known library classes. If there is some java based oss-fuzz project that generates a large list of dependencies and include them as jar files in the output directory, those dependencies are also analysed by fuzz-introspector if their package name is not specifically excluded. This setting actually creates a large overhead by analysing classes which are not part of the target project itself. Indeed, may project runs into OOM or timeout during fuzz-introspector processing because of this reason. This PR aims to decrease the overhead by allowing the project specifying its package prefix and fuzz-introspector will only go through classes with the specified package prefix. If no package prefix has been specified, then it will work like before, in which the fuzz-introspector will analyse all classes in all jar files. This setting is used to be backward compatible for projects which do not specify the package prefix.
In order to apply the new package prefix specification, a special comment line must add to the fuzzer executable when they are created in the oss-fuzz build.sh of that project. The special comment are
##PACKAGE_NAME##[Target package prefix, separated by ':']##PACKAGE_NAME##
For example, ##PACKAGE_NAME##org.fuzz-introspector.*:org.oss-fuzz.*##PACKAGE_NAME##

### DIFF
--- a/frontends/java/oss-fuzz-main.py
+++ b/frontends/java/oss-fuzz-main.py
@@ -72,7 +72,14 @@ def find_fuzz_targets(path):
 
       # If wrapper script exists, retrieve the java class with package name
       classname = content.split("--target_class=")[1].split(" ")[0]
-      targets.append(classname)
+
+      # If wrapper script exists, retrieve target analysing package name if exists
+      # If ##PACKAGE_NAME##[package]##PACKAGE_NAME## exists, java frontend will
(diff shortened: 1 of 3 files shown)


arthurscchan:
Wrong push.

PR #951, opened 2023-03-30
Fix bug: Fix parameter for frontend call

arthurscchan:
The parameter list for calling the java frontend in oss-fuzz-main is in wrong order after adding a new parameter. This PR fixes it by changing the parameter list to correct order.


DavidKorczynski (approved):
LGTM

Down the line we should do so the ordering doesn't matter and have named arguments instead.

arthurscchan:
Sure

PR #1012, opened 2023-05-01
Optional language param for print_function_details

sg3-141-592:
I ran into an issue running some tests with the introspector tool `function_inspector` `print_function_details`. I traced the issue down to a problem in `run_fuzz_introspector_on_dir` where `IntrospectionProject` is always being called with language `c-cpp`. When this is called with `python` instead it runs correctly.

I've added an optional language param. Try

``` python
# Doesn't work
import function_inspector

function_inspector.print_function_details("croniter", ["croniter.croniter.croniter.get_next"])

# Now works
function_inspector.print_function_details("croniter", ["croniter.croniter.croniter.get_next"], language="python")
```

I also added a type hint to the `functions_to_analyse` param as it look me a while to realise it only accepts a list of strings.

### DIFF
--- a/tools/oss-fuzz-scanner/function_inspector.py
+++ b/tools/oss-fuzz-scanner/function_inspector.py
@@ -15,10 +15,12 @@
 
 import sys
 import scanner
+import typing
 
 
-def print_function_details(project_name, functions_to_analyse):
-    report_generator = scanner.get_all_reports([project_name], 2)
+def print_function_details(project_name, functions_to_analyse: typing.List[str],
+    language: str = 'c-cpp'):
+    report_generator = scanner.get_all_reports([project_name], 2, language=language)
 
     project, date_as_str, introspector_project = next(report_generator)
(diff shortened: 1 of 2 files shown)


tools/oss-fuzz-scanner/function_inspector.py
```diff
@@ -43,8 +45,13 @@ def print_function_details(project_name, functions_to_analyse):
...
+            "Usage: python3 function_inspector.py project_name function_name language")
         sys.exit(0)
     project_name = sys.argv[1]
     function_name = sys.argv[2]
-    print_function_details(project_name, [function_name])
+    language = None
+    try:
+        language = sys.argv[3]
+    except IndexError:
+        pass
```
DavidKorczynski, line 56: could we have `language='c-cpp'` here just to avoid passing `language=None` into `print_function_details`, which I assume will produce an error?

DavidKorczynski (changes requested):
Great, thanks!

DavidKorczynski (approved):

PR #1055, opened 2023-06-07
Java-frontend: Fix constructor ignore option

arthurscchan:
Some oss-fuzz project have fuzzers like https://github.com/google/oss-fuzz/blob/master/projects/jettison/JsonFuzzer.java which only call constructor of the target project. There is also some project logic are initialised from constructor. That cause problems to Java-frontend static analysis of these project because the java frontend ignore constructor when generating call graphs by default. The original reason for that is to avoid some large project call graph to be flooding with not useful data. This PR fixes that by adding options for user to include what methods to be excluded during the frontend process. By default, general constructor will not be ignored.


DavidKorczynski:
Could you rebase?

DavidKorczynski (approved):

PR #1085, opened 2023-06-20
Auto-fuzz: Add benchmark summary to post processing

arthurscchan:
Add additional function to post processing to display a list of the target method and which generated fuzzer does cover that method and run successfully. Also, a list of redundant methods covered by successful running fuzzers are also listed to demonstrate what methods should be filtered out. That list should be empty if Auto-fuzz does the filtering correctly and complete. This PR also fixes the README to describe the newly added function.


DavidKorczynski (commented):
The 8 benchmarks are still considered one project though. This is limiting as it's not easy to distinguish data produced by auto-fuzz on a benchmark-level which we need

DavidKorczynski (approved):

PR #1088, opened 2023-06-21
Auto-fuzz: Filter method

arthurscchan:
This PR fixes the frontend to allow a fully-qualified name stored for each method callsite. Then the callsite list of all target methods are retrieve to indicate which methods have already been covered by other target methods and thus filtering them out.


DavidKorczynski (approved):

DavidKorczynski (approved):
Do you have some stats on how much reduction happens for some targets? If so, could you leave it in the comments for reference

arthurscchan:
For real project testing on xz-java, decreased from 7XX fuzzers to 59 fuzzers.

DavidKorczynski:
> For real project testing on xz-java, decreased from 7XX fuzzers to 59 fuzzers.

Nice! Did it exclude fuzzers that were of importance?

PR #1101, opened 2023-06-26
Java-frontend: Fix class filtering in frontend

arthurscchan:
Some projects may include too much dependency libraries in their packaged jar files. This setting affects the effectivenes of frontend analysis on the project code as there are too much noise. This PR adds a feature to the java-frontend for filtering out classes that are not belongs to the target project. If the source directory of the target project is provided, the Java frontend will scan through that directory for a list of Java files. If a class is not in that list, it is skipped during the frontend processing.

### DIFF
--- a/frontends/java/oss-fuzz-main.py
+++ b/frontends/java/oss-fuzz-main.py
@@ -124,6 +124,7 @@ def run_introspector_frontend(target_class, jar_set):
       "fuzzerTestOneInput", # entry method
       package_name, # target package prefix
       "\"<clinit>:finalize\"", # exclude method list
+      "NULL", # source directory
       "False", # Auto-fuzz switch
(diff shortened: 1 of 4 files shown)


arthurscchan:
This PR adds a new parameter to the frontend CallGraphGenerator. Default value has been set for both oss-fuzz-manager and run.sh of the java frontend. 
For general oss-fuzz setup, the same default value are used in the oss-fuzz-manager, so this PR does not affect the oss-fuzz setup.
This PR does affect the auto-fuzz setup or other execution of the Java frontend through run.sh script. If the new --src parameter is provided, it will only process the classes exist in that directory. Otherwise the behaviour is same as before.

DavidKorczynski (approved):

PR #1104, opened 2023-06-26
Auto-fuzz: Fix logic for call depth filtering

arthurscchan:
This PR fixes the logic for call depth filtering and change the default ratio after source folder filtering has been added to the frontend processing in #1101.


arthurscchan:
This PR needs rebase after #1103 is merged in.

DavidKorczynski (approved):

PR #1110, opened 2023-06-27
Auto-fuzz: Ignore unrelated jar

arthurscchan:
Some jar file in the project directory may include some unsolved dependencies, they should be solved and included in the build.jar. Thus there is no need to include them for the fuzzer compile process. This PR fixes the discovery logic of jar files and remove the possibility of compilation error on those unrelated jar.


arthurscchan:
This PR also ignores sources.jar and javadoc.jar which contains part of the source code and won't be able to compile because of missing dependencies. These two jar files are generated only for reference and documentations. The compiled logic of the project will all located in the general jar file.

DavidKorczynski (approved):

PR #1113, opened 2023-06-27
Auto-fuzz: Fix filtering logic for method call depth

arthurscchan:
This PR modifies the method call depth filtering logic. The new logic sort the target method list in descending order by the method call depth and only choose the top 20 methods as target.


arthurscchan:
This PR needs rebase after #1112 is merged.

DavidKorczynski (approved):

PR #1114, opened 2023-06-27
Auto-fuzz: Fix factory method discovery

arthurscchan:
For some of the non-static methods, object creation is required. In the current logic, one of the way for object creation is to search for factory method and call them. The current logic of factory method discovery create too much noise. This PR fixes the logic of the factory method discovery to decrease noise and hence decrease the number of generated fuzzers with incorrect factory methods.


arthurscchan:
This PR needs rebase after #1113 is merged in.

DavidKorczynski (approved):

PR #1126, opened 2023-06-29
Auto-fuzz: Exclude some dependencies for java frontend

arthurscchan:
Some project requires a long list of dependencies for using their code. This long list of dependencies with flood the java frontend and may cause failing because of out of memory and time out. This PR fixes the logic of project build to move the dependency jar into a separate directory which could then be excluded in the Java frontend process.


DavidKorczynski (approved):

DavidKorczynski:
Could you fix codestyle?

PR #1142, opened 2023-07-06
Auto-fuzz: Fix gradle build skip

arthurscchan:
This PR fixes the spotlessCheck skip bug when spotlessCheck task is not defined in the project.


DavidKorczynski:
I assume this is a follow up to ? https://github.com/ossf/fuzz-introspector/pull/1140

DavidKorczynski (approved):

PR #1155, opened 2023-07-27
Java-frontend: Fix method filtering for auto-fuzz

arthurscchan:
This PR fixes the method and class filtering logic to reduce the output data.yaml files for auto-fuzz generation.


DavidKorczynski (approved):

DavidKorczynski:
Just to clarify, what is the impact of this in non-auto-fuzz mode?

PR #1167, opened 2023-07-31
Auto-fuzz: Fix constructor filtering

arthurscchan:
This PR adds additional filtering for constructors to exclude non-project class constructor in resulting fuzzers.


DavidKorczynski (changes requested):
Could you fix the formatting?

DavidKorczynski (approved):

Issue #1174, opened 2023-08-01
Fuzz-introspector fails for projects in oss-fuzz with "'c' flag requires manifest or input files to be specified!"


centic9:
The java frontend currently fails for a number of projects in oss-fuzz with a message `'c' flag requires manifest or input files to be specified!`. 

See the recent introspection builds at e.g. https://oss-fuzz-build-logs.storage.googleapis.com/index.html#apache-commons-bcel or https://oss-fuzz-build-logs.storage.googleapis.com/index.html#apache-poi (among likely other projects with similar errors).

```
Step #6 - "compile-libfuzzer-introspector-x86_64": Finding fuzz targets in /workspace/out/libfuzzer-introspector-x86_64
Step #6 - "compile-libfuzzer-introspector-x86_64": Checking target textcov_reports
Step #6 - "compile-libfuzzer-introspector-x86_64": Checking target jazzer_driver_with_sanitizer
Step #6 - "compile-libfuzzer-introspector-x86_64": Checking target poi-fuzzer-5.2.4-SNAPSHOT.jar
Step #6 - "compile-libfuzzer-introspector-x86_64": Checking target XLSX2CSVFuzzer
Step #6 - "compile-libfuzzer-introspector-x86_64": Checking target llvm-symbolizer
Step #6 - "compile-libfuzzer-introspector-x86_64": Checking target jazzer_driver
Step #6 - "compile-libfuzzer-introspector-x86_64": Checking target jazzer_agent_deploy.jar
Step #6 - "compile-libfuzzer-introspector-x86_64": 'c' flag requires manifest or input files to be specified!
Step #6 - "compile-libfuzzer-introspector-x86_64": Try `jar --help' for more information.
Step #6 - "compile-libfuzzer-introspector-x86_64": Traceback (most recent call last):
Step #6 - "compile-libfuzzer-introspector-x86_64":   File "/fuzz-introspector/frontends/java/oss-fuzz-main.py", line 168, in <module>
Step #6 - "compile-libfuzzer-introspector-x86_64":     run_analysis(os.environ['OUT'])
Step #6 - "compile-libfuzzer-introspector-x86_64":   File "/fuzz-introspector/frontends/java/oss-fuzz-main.py", line 154, in run_analysis
Step #6 - "compile-libfuzzer-introspector-x86_64":     targets = find_fuzz_targets(path)
[...]

PR #1176, opened 2023-08-01
Fix empty class file list bug

arthurscchan:
As mentioned in #1174, some projects do not have any loose class files to package because of different reasons. Thus passing an empty list of class files to the jar tools result in errors in shell execution. This PR fixes that by adding a conditional check to only run the jar tools when any loose class files are discovered. It will not affect the frontend execution as jar files packaged by the project itself would still be discovered in later code.

### DIFF
--- a/frontends/java/oss-fuzz-main.py
+++ b/frontends/java/oss-fuzz-main.py
@@ -87,8 +87,9 @@ def find_fuzz_targets(path):
               os.rename(full_path, os.path.join(path, class_location))
             class_file_list.append(class_location)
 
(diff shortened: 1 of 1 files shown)


DavidKorczynski:
@centic9 could you help review?

centic9 (approved):
LGTM, 

We could add a line `print("Found fuzz-target %s" % classname)` after `targets.append(classname)` to allow to see which targets are actually found at this point.

centic9 (approved):

centic9 (approved):
LGTM

PR #1180, opened 2023-08-03
tools: add a top level extractor

DavidKorczynski:


AdamKorcz (approved):

oliverchang:
Thanks for this! There's a bug here that we need to handle though, which is that we should ignore "top level" functions defined in the same fuzz target file. 

e.g. https://github.com/google/oss-fuzz/blob/102a2017691072b5acd91c18a72bd0a024cb1ce2/projects/elfutils/fuzz-libelf.c

Currently the "top level functions" are being reported as  "fuzz_logic_one"

Issue #1188, opened 2023-08-07
web: bool types are reported as "N/A" ? 


oliverchang:
e.g. https://github.com/vstakhov/libucl/blob/c8374381bab04f33f0000910acdf37593e343c99/include/ucl.h#L501

https://introspector.oss-fuzz.com/api/far-reach-but-low-coverage?project=libucl says:

`"function-arguments":"['struct.ucl_object_s *', 'struct.ucl_object_s *', 'N/A']","function-name":"ucl_array_merge"`

oliverchang:
CC @DavidKorczynski

PR #1190, opened 2023-08-07
frontends: llvm: extend type recognition

DavidKorczynski:
Extends to recognizing bool, float, double, short and void.


DavidKorczynski:
Ref https://github.com/ossf/fuzz-introspector/issues/1188

AdamKorcz (approved):

PR #1201, opened 2023-08-21
doc: installation doc fixes

fanquake:
Minor changes after running into issues following the installation instructions here: https://fuzz-introspector.readthedocs.io/en/latest/installation-instructions/installation.html. Using the main branch of LLVM doesn't work, as much has changed, and the patches don't apply. Given OSS-Fuzz uses the LLVM 15 branch, it seemed reasonable to pin to that (can also drop to 14). LLVM 15 requires a `DCMAKE_BUILD_TYPE` to be specificed when configuring, given that 14 defaulted to `Debug` if unset, I've used that here. Also fix the name of the patch-file.

### DIFF
--- a/doc/LocalBuild.md
+++ b/doc/LocalBuild.md
@@ -65,6 +65,7 @@ cd ../
 # Build LLVM and Clang
 git clone https://github.com/llvm/llvm-project/
 cd llvm-project/
+git checkout release/15.x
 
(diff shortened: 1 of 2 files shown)


DavidKorczynski (approved):
Thanks a ton!

I didn't run this through testing yet but will do so later.

Issue #585, opened 2022-11-05
Performance improvements in post-processing


Navidem:
Some projects like `openssl` or `binutils` are hitting the generous timeout of 16hrs on cloud build. This is while `openssl` finishes locally within 6hrs. 
We may start looking at opportunities to improve performance in python post-processing.

`openssl` log for Nov 3rd: [log-954de748-0033-46cb-9eb1-bb78f4013819.txt](https://github.com/ossf/fuzz-introspector/files/9942682/log-954de748-0033-46cb-9eb1-bb78f4013819.txt)

maxammann:
Do we have a log file which includes timestamps, so we can better see where the time is spent?

DavidKorczynski:
> Do we have a log file which includes timestamps, so we can better see where the time is spent?

I added timestamps to the logger used for output a few days ago: https://github.com/ossf/fuzz-introspector/pull/1210

As such, it should be available if you use the latest, e.g.:

```
git clone https://github.com/ossf/fuzz-introspector
cd fuzz-introspector/oss_fuzz_integration
./build_post_processing.sh
cd oss-fuzz
python3 infra/helper.py build_fuzzers --sanitizer=introspector htslib
```

Should give you timestamps in the output.

PR #1203, opened 2023-08-22
frontend: llvm: avoid writing multiple YAML modules

DavidKorczynski:
This was initially introduced to solve memory exhaustion issues. But I believe the memory exhaustion issues were actually solved in https://github.com/ossf/fuzz-introspector/pull/561

Keeping things in one module should speed up the proces, which we have had issues with on-and-off, such as in openssl and binutils: https://github.com/ossf/fuzz-introspector/issues/585


DavidKorczynski:
@jonathanmetzman fyi

DavidKorczynski:
This will help both when reating the .yamls, but, most importantly when loading in the yaml modules in the post processing here: https://github.com/ossf/fuzz-introspector/blob/302d36bf21ee19ea97d4742711691ed3ba692a31/src/fuzz_introspector/utils.py#L76-L133

AdamKorcz (approved):

jonathanmetzman:
Thanks!

PR #1205, opened 2023-08-23
Java-frontend: Fix entry method discovery

arthurscchan:
It is discovered that Jazzer allows using custom methods with FuzzTest annotation to indicate the method to be fuzzed in addition to the default fuzzerTestOneInput method. The current JVM frontend logic fails to discover these custom methods and causes some projects using the annotation approach to fail to generate fuzzer profile for fuzz-introspector build. This PR fixes the frontend to add in annotation scanning to look for possible fuzzing methods if fuzzerTestOneInput is not found in the fuzzer class.
Example project: https://github.com/google/oss-fuzz/tree/master/projects/apache-cxf

### DIFF
--- a/frontends/java/src/main/java/ossf/fuzz/introspector/soot/CallGraphGenerator.java
+++ b/frontends/java/src/main/java/ossf/fuzz/introspector/soot/CallGraphGenerator.java
@@ -52,8 +52,6 @@ import soot.SceneTransformer;
 import soot.SootClass;
 import soot.SootField;
 import soot.SootMethod;
-import soot.tagkit.AnnotationTag;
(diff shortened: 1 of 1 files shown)


DavidKorczynski (changes requested):
Just for reference, could you link to an example that uses this set up?

DavidKorczynski:
Example fuzzer: https://github.com/google/oss-fuzz/blob/7e4fde504284aeaca6094cabe555521786d12448/projects/apache-cxf/project-parent/fuzz-targets/src/test/java/com/example/AttachmentSerializerDeserializerFuzzer.java#L45-L46

DavidKorczynski (approved):

PR #1208, opened 2023-08-24
frontend: java, multithread

DavidKorczynski:


DavidKorczynski:
This does not seem to work because SOOT does some locking on the java files it analyses. Performance gains was found elsewhere.

PR #1229, opened 2023-08-31
webapp: display language build stats on UI and update readme

DavidKorczynski:


DavidKorczynski:
Will add some UI as well to display it rather than just logging

AdamKorcz (approved):

PR #1239, opened 2023-09-21
JVM: Fix constructor display for jvm project

arthurscchan:
The current logic fails to display the name of the constructor for jvm project because the method name returned by Soot is <init> or <cinit> for class constructors and browsers treat it as a general html tag. This PR adds additional logic to encode the < and > character to its entity name &lt; and &gt;  for correct display in the html.

### DIFF
--- a/src/fuzz_introspector/analyses/calltree_analysis.py
+++ b/src/fuzz_introspector/analyses/calltree_analysis.py
@@ -15,6 +15,7 @@
 
 import os
 import logging
+import html
 import json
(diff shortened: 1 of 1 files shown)


DavidKorczynski (approved):

DavidKorczynski (changes requested):
We should watch out if this changes things other than HTML displayed text, which I think it does

DavidKorczynski (approved):

PR #1248, opened 2023-10-09
Auto-fuzz: avoid rebuild project

arthurscchan:
In the current auto-fuzz logic, all the successfully generated fuzzer integration are copied to oss-fuzz for test build and test run. This process takes up the most time for the auto-fuzz generation because the target project needs to be built for every integration. This PR changes the logic to reuse built jar files from the first project build to avoid further attempts at project build. It then replaces the docker file and build.sh with the simplified version with a specific JDK version and project build type. As a result, this PR decrease the total overhead and execution time of the auto-fuzz generation and testing process.

### DIFF
--- a/tools/auto-fuzz/base_files.py
+++ b/tools/auto-fuzz/base_files.py
@@ -57,7 +57,8 @@ def gen_dockerfile(github_url,
     if language == "python":
         return gen_dockerfile_python(github_url, project_name)
     elif language == "jvm":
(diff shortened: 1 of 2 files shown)


DavidKorczynski:
Is it possible to come up with a more generic solution so it will also work for Python?

arthurscchan:
Not sure about that. But I don't think python could do the similar. Cause python project is not really "building", it is just installing the package. And I don't think python project needed to be built at all. WDYT??

PR #1252, opened 2023-10-10
Auto-fuzz: Refine base file

arthurscchan:
Refine the logic of the base project file retrieval and the logic for detecting of project build type.


arthurscchan:
Wrong implementation.

PR #1253, opened 2023-10-10
Auto-fuzz: Extract base file templates and add project language and build type detection logic

arthurscchan:
This PR changes the logic for base project file generation and detection of project build type. Templates for those base project has been extracted to the template directory. The new logic will detect the language and project build type of the target projects. It will use the information to retrieve and generate correct base project files which also configure the JDK versions.


DavidKorczynski (changes requested):
Could you add a bit more description in the PR on what this changes?

Also, does `project_type` only reflect which build tool is used? If so, then I think we should adjust the name to something related to `build`.

tools/auto-fuzz/manager.py
```diff
@@ -135,27 +135,37 @@ def project_name(self):
...
-    def write_basefiles(self):
+    def write_basefiles(self, project_build_type=None):
         with open(self.build_script, "w") as bfile:
-            bfile.write(base_files.gen_builder_1(self.language))
+            bfile.write(
+                base_files.gen_builder_1(self.language, project_build_type))
 
         with open(self.base_fuzzer, "w") as ffile:
-            ffile.write(base_files.gen_base_fuzzer(self.language))
+            ffile.write(
```
DavidKorczynski, line 144: Could we use a better name than `ffile`?

DavidKorczynski (changes requested):

tools/auto-fuzz/manager.py
arthurscchan: Changed..

DavidKorczynski (approved):
LGTM, nice refactoring!

PR #1254, opened 2023-10-10
Auto-fuzz: Adopt OSS-Fuzz for project build and avoid multiple build for the same project

arthurscchan:
This PR changes the logic for project building and static analysis before the fuzzer generation process. The new logic removes the project-building process from manager.py. Instead, it detects the language and project build type at the beginning and generates the base-autofuzz directory. The generated base-autofuzz directory is then passed to oss-fuzz for project building and save all the build-jar files for later use. This makes all the build project make use of the existing oss-fuzz settings and avoid the need to rebuild the process when testing each of the generated fuzzer integrations.

The old flow for the auto-fuzz manager.py
1. **Build the target project in the local environment.**
2. **Run static analysis using the local build of the target project.**
3. Create the necessary files for the base autofuzz directory
4. Copy the base autofuzz directory to build and test it.
5. Use the static analysis result for fuzzing target generation
6. Generate separate fuzzing integration directories and test them one by one.

The new flow for the auto-fuzz manager.py
1. **Detecting the project build type of the target project.**
2. **Creating the base autofuzz directory with the correct language and project build type**
3. **Copy the base autofuzz directory to oss-fuzz and build it.**
4. **Retrieved the built jar files and replace the Dockerfile and build script of the base autofuzz directory to skip re-building of the process and add in code to copy the build jar files.**
5. Run static analysis using the oss-fuzz build of the target project.
6. Create the necessary files for the base autofuzz directory
7. Copy the base autofuzz directory to build and test it.
8. Use the static analysis result for fuzzing target generation
9. Generate separate fuzzing integration directories and test them one by one (with the abstracted build script and dockerfile which skip the project build and make use of the build jar files).
10. **Replace the dockerfile and the build script of the project with the normal one after the testing.**

With this PR, the build script has been significantly decreased by just concentrating on a single project build type. Also, the testing process is significantly increased because the target project has only been built once and all the build jar files are reused. For illustration, the project `javassist` has 148 fuzzing targets generated. The original testing time for all targets is around 45 mins. The new logic takes around 17mins.

### DIFF
--- a/tools/auto-fuzz/base_files.py
+++ b/tools/auto-fuzz/base_files.py
@@ -125,9 +125,9 @@ def _gen_builder_1_jvm(template_dir, build_project):
         BASE_BUILDER = "#!/bin/bash -eu\n" + file.read()
 
     if build_project:
-        return BASE_BUILDER % ("", "", ": << COMMENT", "COMMENT")
+        return BASE_BUILDER % ("", "", ": <<'COMMENT'", "COMMENT")
     else:
-        return BASE_BUILDER % (": << COMMENT", "COMMENT", "", "")
+        return BASE_BUILDER % (": <<'COMMENT'", "COMMENT", "", "")
 
 
 def _gen_base_fuzzer_python(template_dir):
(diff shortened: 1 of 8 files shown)


DavidKorczynski:
> The original testing time for all targets is around 45 mins. The new logic takes around 17mins.

Nice, exact same results in terms of fuzzers generated as well?

DavidKorczynski (approved):

PR #1255, opened 2023-10-11
Auto-fuzz: Extraction of classes and functions from manager.py

arthurscchan:
This PR tidies up the code for manager.py by extracting those util functions and the oss-fuzz-project class into separate Python files to make manager.py less crowded and only control auto-fuzz flow related functions.


DavidKorczynski (changes requested):
Could we change the name from classes.py to something different? I think it's too close to `class` which is a keyword in Python.

DavidKorczynski (approved):

PR #1265, opened 2023-10-16
Auto-fuzz: Add oss-fuzz docker support for java static analysis

arthurscchan:
This PR changes the logic for static analysis of the Java project. Instead of running the static analysis in the local environment, the new logic builds an OSS-Fuzz directory to run the static analysis for the Java project in the OSS-Fuzz docker image. This allows more flexibility in the settings and dependencies that are required for the Java static analysis without the need to load and depend on the settings in the local environment. The next step is to make this part generalize for all languages.

### DIFF
--- a/tools/auto-fuzz/templates/java-introspector/Dockerfile-template
+++ b/tools/auto-fuzz/templates/java-introspector/Dockerfile-template
@@ -1,0 +1,29 @@
+# Copyright 2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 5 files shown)


tools/auto-fuzz/templates/java-introspector/Dockerfile-template
```diff
@@ -0,0 +1,29 @@
...
+#
+##########################################################################
+FROM gcr.io/oss-fuzz-base/base-builder-jvm
+#RUN curl -L %s -o maven.zip && unzip maven.zip -d $SRC/maven && rm -rf maven.zip
+RUN curl -L %s -o jdk.tar.gz && tar zxf jdk.tar.gz && rm -rf jdk.tar.gz
+COPY maven.zip $SRC/maven.zip
+RUN unzip maven.zip -d $SRC/maven && rm ./maven.zip
+ENV JAVA_HOME="$SRC/%s"
+ENV PATH="$JAVA_HOME/bin:$SRC/maven/apache-maven-3.6.3/bin:$PATH"
+RUN git clone --depth 1 https://github.com/ossf/fuzz-introspector introspector
```
DavidKorczynski, line 23: Can we switch this so it uses the local version that exists for `auto-fuzz`? THat makes it easier to develop I assume

arthurscchan: done.

DavidKorczynski (approved):

PR #1306, opened 2023-11-13
Auto-fuzz: Create dockerfile for fuzzer generator

arthurscchan:
This PR creates a dockerfile for building a custom docker image. The custom docker image will be used for the fuzzer generator for different language to avoid the need to installed specific python package.

### DIFF
--- a/tools/auto-fuzz/fuzzer_generator/docker/Dockerfile
+++ b/tools/auto-fuzz/fuzzer_generator/docker/Dockerfile
@@ -11,23 +11,24 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+##########################################################################
+FROM ubuntu:22.04@sha256:2b7412e6465c3c7fc5bb21d3e6f1917c167358449fecac8176c6e496e5c1f05f
 
-FROM ubuntu:22.04
+RUN apt-get update && \
+    apt-get install pip git -y
 
-RUN apt-get update && apt-get install git pip curl -y
+ENV SRC /src
+ENV OUT /out
 
-RUN curl -sSL https://get.docker.com/ | sh
-RUN git clone https://github.com/ossf/fuzz-introspector fuzz-introspector
+RUN mkdir -p $SRC && \
+    mkdir -p $OUT
 
-WORKDIR $SRC/fuzz-introspector
-RUN git submodule init
-RUN git submodule update
+COPY fuzz-introspector $SRC/fuzz-introspector
+COPY work $SRC/work
 
-RUN pip3 install --upgrade pip
-RUN pip3 install -r ./requirements.txt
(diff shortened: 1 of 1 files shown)


tools/auto-fuzz/fuzzer_generator/docker/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+FROM ubuntu:22.04
+
+RUN apt-get update && apt-get install git pip curl -y
+
+RUN curl -sSL https://get.docker.com/ | sh
+RUN git clone https://github.com/ossf/fuzz-introspector fuzz-introspector
```
DavidKorczynski, line 20: Could we use the local copy of fuzz-introspector instead? I assume at least for now we would want to prefer a set up that makes it easy to develop on

DavidKorczynski (changes requested):

tools/auto-fuzz/fuzzer_generator/docker/Dockerfile
arthurscchan: Changed.

DavidKorczynski (changes requested):
Could you give some explanation on how to use this?

tools/auto-fuzz/fuzzer_generator/docker/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+RUN mkdir -p $OUT
+
+RUN apt-get update && apt-get install pip git -y
+
+COPY fuzz-introspector work $SRC/
+
+WORKDIR $SRC/fuzz-introspector
+RUN git submodule init
+RUN git submodule update
+RUN pip3 install -r $SRC/requirements.txt
```
DavidKorczynski, line 30: Could you just do all of the four above commands in a single `RUN` docker command?

DavidKorczynski (changes requested):

tools/auto-fuzz/fuzzer_generator/docker/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+
+RUN mkdir -p $SRC
+RUN mkdir -p $OUT
+
+RUN apt-get update && apt-get install pip git -y
+
+COPY fuzz-introspector work $SRC/
+
+WORKDIR $SRC/fuzz-introspector
+RUN git submodule init && git submodule update && pip3 install -r ./requirements.txt
```
DavidKorczynski, line 28: please do a line per cmd as oss-fuzz Dockerfiles https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-builder/Dockerfile

DavidKorczynski (changes requested):

DavidKorczynski (approved):

PR #1310, opened 2023-11-13
Auto-fuzz: Build docker image for fuzzer generation

arthurscchan:
This PR follows #1306 which includes logic to build the docker image for the fuzzer generation process and run a simple ls command in a docker container of the built image. The addition of the logic does not affect the process of manager.py. There will be another PR later to change the flow of manager to use the docker container to do the fuzzer generation process.

This PR needs rebase after the docker file created in #1306 has been merged.


arthurscchan:
This PR is closed because the approach has been changed. New PR is coming in later.

PR #1344, opened 2024-01-08
json-reporting: add callsites dictionary to output

DavidKorczynski:
We have callsites for each function with respect to what functions they are calling. However, it's currently not output to `summary.json`. This fixes it. 

Ref: https://github.com/ossf/fuzz-introspector/issues/1343


DavidKorczynski:
CFLite issues are false positives.

PR #1347, opened 2024-01-09
Sink analyser: Extend sink analyser for more cwe

arthurscchan:
This PR extends the sink analyser to support more than one CWE and refine the logic for analysing sink function existence. This is the first PR to update the sink analyser to a larger CWE analyser.


src/fuzz_introspector/analyses/data/cwe_data.py
```diff
@@ -0,0 +1,119 @@
+# Copyright 2023 Fuzz Introspector Authors
```
DavidKorczynski, line 1: 2024

src/fuzz_introspector/analyses/data/cwe_data.py
```diff
@@ -0,0 +1,119 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Any, Dict
+
+SINK_FUNCTION: Dict[str, Dict[str, Any]] = {
+    'CWE79': {
```
DavidKorczynski, line 18: How about making this a bit more typed than a dictionary? It would be useful to have some form of metadata related to each CWE I think

DavidKorczynski (changes requested):

DavidKorczynski (approved):

PR #1444, opened 2024-02-25
frontends: llvm: add structured debug information output

DavidKorczynski:


DavidKorczynski:
Will also add post-processing logic for now, as we should phase out the custom parsing that is happening.

AdamKorcz (approved):

PR #1468, opened 2024-03-11
webapp: provide initial sample of using type map

DavidKorczynski:


DavidKorczynski:
We need to hold on to this for a while, because storing it inside the webapp causes it too take up too much space.

PR #1474, opened 2024-03-14
Update all-cross-references

trashvisor:
Return error when a function signature can't be found.


DavidKorczynski (changes requested):
Could you do a quick `yapf -i` on `routes.py`?

DavidKorczynski (approved):

PR #1540, opened 2024-06-03
separate all function list out from `summary.json`

DavidKorczynski:
This will improve the speed of creating the DB in the webapp significantly, since we don't always have to download large summary.json files.


DavidKorczynski:
CFLite is a false positive. This s good for review

PR #1558, opened 2024-06-11
[webapp] Add new api for filtered functions

arthurscchan:
This PR adds two additional API for querying the all-functions and all-jvm-constructors list with filtering to ensure only publicly accessible and non-jvm library functions are included in the list. This only works for JVM projects since all other project functions is_public and is_jvm_library field is always default to true and false respectively which passes the filter.


tools/web-fuzzing-introspection/app/webapp/helper/function_helper.py
```diff
@@ -0,0 +1,76 @@
...
+    else:
+        functions = [
+            target for target in target_list if target.project == project_name
+        ]
+
+    return _convert_functions(_sort_functions(functions))
+
+
+def _sort_functions(functions: List[models.Function]) -> List[models.Function]:
+    """Sort the function list according to certain criteria."""
```
DavidKorczynski, line 46: Could we make this more precise? `certain criteria` is too general

tools/web-fuzzing-introspection/app/webapp/helper/function_helper.py
```diff
@@ -0,0 +1,76 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import sys
+from typing import Dict, List, Any
+
+from .. import models
+
+
+def process_functions(target_list: List[models.Function], project_name: str,
```
DavidKorczynski, line 21: Could we make this more precise in the name? A lot of the logic in the web app is about "processing functions"

tools/web-fuzzing-introspection/app/webapp/helper/function_helper.py
```diff
@@ -0,0 +1,76 @@
...
+            'raw_function_name': function.raw_function_name,
+            'is_reached': function.is_reached,
+            'accummulated_complexity':
+            function.accummulated_cyclomatic_complexity,
+            'function_argument_names': function.function_argument_names,
+            'function_arguments': function.function_arguments,
+            'reached_by_fuzzers': function.reached_by_fuzzers,
+            'return_type': function.return_type,
+            'runtime_coverage_percent': function.runtime_code_coverage,
+            'is_enum_class': function.is_enum_class
```
DavidKorczynski, line 74: What's the idea behind returning all of this information? Where will it be processed?

DavidKorczynski (changes requested):

tools/web-fuzzing-introspection/app/webapp/helper/function_helper.py
arthurscchan: changed.

tools/web-fuzzing-introspection/app/webapp/helper/function_helper.py
arthurscchan: It is not only used for the new oracle, it is also used for the original all-functions and all-jvm-constructors api which return this list of information. It is used here only to preserve the functionality of the all-functions and all-jvm-constructors api and to generalise the return.

DavidKorczynski (changes requested):

DavidKorczynski (approved):

PR #1568, opened 2024-06-12
Fix wrong overwrite of __dict__ attribute

arthurscchan:
This PR fixes the wrong overwrite for the __dict__ attribute. It is an attribute instead of function.


DavidKorczynski (changes requested):
This will break here: https://github.com/ossf/fuzz-introspector/blob/ef663764de877608f56766ddf0be479c4bdaf521/tools/web-fuzzing-introspection/app/webapp/routes.py#L1504

DavidKorczynski (approved):

PR #1572, opened 2024-06-12
CI: Add new CI for testing webapp API

arthurscchan:
This PR adds a new CI to test the webapp API and see if there is unexpected error occur.


.github/helper/api_test.py
```diff
@@ -0,0 +1,67 @@
...
+  ROOT_FI = os.path.join(os.path.dirname(os.path.abspath(__file__)), '..', '..')
+
+  code = subprocess.call([os.path.join(ROOT_FI, '.github', 'helper', 'prepare_webapp')])
+  if code:
+    EXCEPTIONS.append('Failed to prepare the webapp for testing.')
+
+  # A list of curl test to the webapp api
+  for project in PROJECTS:
+    for api in APIS:
+      _test_server_api(f'{BASE_URL}{api}?project={project}')
```
DavidKorczynski, line 57: We need better parameters here as the current set up won't trigger most of the code. We should set up it up such that we can specify an API as well as the arguments it takes.

DavidKorczynski (changes requested):

.github/helper/api_test.py
DavidKorczynski: Just noticed this is draft -- I assume it's probably in draft because this is WIP. Will resolve this and re-review after.

DavidKorczynski (approved):

PR #1576, opened 2024-06-13
webapp: Fix NoneType bug

arthurscchan:
When a project name not existed in the db is provided to /api/far-reach-low-cov-fuzz-keyword, the target_project will be None and that cause NoneType error in later call to function `oracle_1`. This PR fixes that by adding a NoneType check.


DavidKorczynski:
Fix already in https://github.com/ossf/fuzz-introspector/pull/1574 -- good to know mypy found it though

PR #1581, opened 2024-06-14
webapp: relocate introspector functions and add typing

arthurscchan:
This PR extracts those introspector related functions to a separate helper and add typing to those functions to aid mypy doing type check.

**This PR is not ready, it is still in progress and require more testing before it is ready for review.**


arthurscchan:
Not needed anymore.

PR #1583, opened 2024-06-14
webapp: Add typing for mypy check

arthurscchan:
This PR aims to add typing for functions and objects in the webapp to allow mypy and pyright CI to detect typing errors.

**This PR is still in progress and not yet ready for review.**

PR #1584, opened 2024-06-14
webapp: Fix type error

arthurscchan:
There are some type errors found in the default values of certain items when typing is added to the webapp functions in #1583. This PR aims to fix it to avoid potential error in type.


DavidKorczynski (approved):
Awesome!

PR #1600, opened 2024-06-17
webapp: Add typing to models.py

arthurscchan:
This PR adds typing to models.py. This PR is one of the smaller PR separated from #1583.

### DIFF
--- a/tools/web-fuzzing-introspection/app/webapp/models.py
+++ b/tools/web-fuzzing-introspection/app/webapp/models.py
@@ -14,59 +14,22 @@
 
 import json
 import datetime
+from typing import Dict, List, Tuple, Any
 
 
-def get_date_at_offset_as_str(day_offset=-1):
+def get_date_at_offset_as_str(day_offset: int = -1) -> str:
     datestr = (datetime.date.today() +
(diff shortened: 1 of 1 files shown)


tools/web-fuzzing-introspection/app/webapp/models.py
```diff
@@ -63,49 +86,33 @@ def has_introspector(self) -> bool:
         return self.introspector_data != None
 
 
-class Project:
```
DavidKorczynski, line 66: For future references, it's a lot easier to review if the location of things are not mved around. For example, the `Project` class is moved from line 66 to line 26, but I don't see a reason why?

arthurscchan: The reason why it is moved up is because the typing for the DBSummary constructor requires Project.
https://github.com/ossf/fuzz-introspector/pull/1600/files#diff-2350a52b5fca0958b98e475be7c02794578f01d9a6748996b9ec6257d3acc28aR62

DavidKorczynski: Ah, makes sense. Thanks!

DavidKorczynski (approved):

PR #1603, opened 2024-06-17
webapp: Ensure None checking for all cases

arthurscchan:
This PR fixes the None type checking for latest_introspector_datestr to ensure it has been checked for all branches. This is part of the work to fulfil the type checking when type is added in https://github.com/ossf/fuzz-introspector/pull/1601.

### DIFF
--- a/tools/web-fuzzing-introspection/app/webapp/routes.py
+++ b/tools/web-fuzzing-introspection/app/webapp/routes.py
@@ -1697,8 +1697,8 @@ def sample_cross_references():
                         if ps.introspector_data is not None:
                             latest_introspector_datestr = datestr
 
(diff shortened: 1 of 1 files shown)


DavidKorczynski (commented):
This seems to be a false positive from MyPy? I don't see how it can be `None` if the first `if` statement is true?

arthurscchan:
Yes, I agree it is a false positive. But changing that does not harm the code at all and it could make mypy go pass. So why not?

DavidKorczynski:
> Yes, I agree it is a false positive. But changing that does not harm the code at all and it could make mypy go pass. So why not?

Fine by me, but I was curious if this was really MyPy reporting it? I thought it was able to catch that this could not be None

PR #1616, opened 2024-06-20
[webapp] Fix possible index error for oracles

arthurscchan:
In #1610, the support for excluding static functions in oracles has been added. The static function checking is supported by the is_static function. The is_static function takes in a source's begin and end line information and invokes the extract_lines_from_source_code function to extract the source for the target function to determine if it is a static function or not. Although the logic has checked if the begin src line is larger than the total source line to avoid index overflow, there is a missing check for a negative begin src line which causes a possible index underflow and results in an Internal Server Error when querying the oracle with `exclude-static-functions=True`
This will happen because in #1555, the webapp has introduced JVM constructors as part of the function candidates, and the source line start information for JVM constructors is always -1 because that information does not exist. Thus it breaks the current logic. 
This PR fixes that by adding a negative checking and returning immediately when a negative src_begin is found to avoid the later index underflow problem.

**Remark: Although the support for JVM constructors has been introduced in #1555, it still does not affect the latest production webapp. The reason is that the JVM constructor information needed for the webapp depends on a JSON file generated by the fuzz-introspector introduced in #1523. This version of the fuzz-introspector has not yet bumped in oss-fuzz. So the constructor list on the webapp production is still empty and hence the bug is not yet appearing on production.**

### DIFF
--- a/tools/web-fuzzing-introspection/app/webapp/routes.py
+++ b/tools/web-fuzzing-introspection/app/webapp/routes.py
@@ -93,12 +93,6 @@ def extract_lines_from_source_code(project_name,
                                    line_end,
                                    print_line_numbers=False,
                                    sanity_check_function_end=False):
-    # Skip the process and return None if src_begin is < 0. This is because
-    # JVM constructors do not have valid source line information and thus
-    # cannot be located in the source file.
-    if line_begin <= 0:
-        return None
-
     raw_source = extract_introspector_raw_source_code(project_name, date_str,
                                                       target_file)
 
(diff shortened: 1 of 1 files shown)


DavidKorczynski (commented):
Can you show the stack trace? I feel like the check should be in `extract_lines_from_source_code` assuming that's where the issue occurs?

DavidKorczynski (changes requested):
We'll still run into issues? it is decreased here so we can end up with -1 : https://github.com/ossf/fuzz-introspector/pull/1616/files#diff-16a502534f1911e9af87997775977329e74d8d9aa3bf52f4b65c5f1e283a594fL107

DavidKorczynski (approved):

PR #1663, opened 2024-07-11
script: Add installation of dependencies to oss-fuzz-gen script

arthurscchan:
It is found that both the webapp and the oss-fuzz-gen require some dependencies that need to be installed. This PR fixes both the run_all.sh and web_run_all.sh by adding dependencies installation in the python venv before starting the webapp or calling oss-fuzz-gen.


DavidKorczynski (changes requested):
Build all does this? https://github.com/ossf/fuzz-introspector/blob/main/scripts/oss-fuzz-gen-e2e/build_all.sh

PR #1667, opened 2024-07-18
core: find and store test files

DavidKorczynski:
First go keeping it as simple as possible

Ref: https://github.com/google/oss-fuzz-gen/issues/494


DavidKorczynski:
CFLite issues is unrelated, this is good to go

AdamKorcz (approved):

PR #1736, opened 2024-09-09
webapp: Fix and unify source file path

arthurscchan:
This PR fixes an issue with non-unified target file strings. For JVM projects, the target file string is a path of the source file within the package directory, which does not start with a '/'. This behaviour differs from other project languages like C/C++, where the path always starts with a '/'. This PR unifies the behaviour to allow the logic to handle both cases successfully.

### DIFF
--- a/tools/web-fuzzing-introspection/app/webapp/routes.py
+++ b/tools/web-fuzzing-introspection/app/webapp/routes.py
@@ -71,8 +71,8 @@ def extract_introspector_raw_source_code(project_name, date_str, target_file):
 
     if is_local:
         src_location = os.path.join(local_oss_fuzz, 'build', 'out',
-                                    project_name, 'inspector',
-                                    'source-code',  target_file)
+                                    project_name, 'inspector', 'source-code',
+                                    target_file)
 
         if not os.path.isfile(src_location):
             return None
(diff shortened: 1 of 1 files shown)


tools/web-fuzzing-introspection/app/webapp/routes.py
```diff
@@ -66,18 +66,23 @@ def get_coverage_report_url(project_name, datestr, language):
...
 def extract_introspector_raw_source_code(project_name, date_str, target_file):
+    if target_file.startswith('/'):
+        target_file = target_file[1:]
 
     if is_local:
         src_location = os.path.join(local_oss_fuzz, 'build', 'out',
-                                    project_name, 'inspector',
-                                    'source-code') + target_file
+                                    project_name, 'inspector', 'source-code',
+                                    target_file)
```
DavidKorczynski, line 75: These changes look quite suspicious -- I don't see how the semantics remain. Why take `+ target_file` into the `os.path.join`?

DavidKorczynski (changes requested):

tools/web-fuzzing-introspection/app/webapp/routes.py
DavidKorczynski: Ah, I think I understand it now. So the whole point is to remove any potential `/` to make the `os.path.join` work

DavidKorczynski (approved):

PR #1739, opened 2024-09-11
Python/JVM: Copy source files for python project

arthurscchan:
This PR fixes the logic of the extract_all_sources function to also handles source files for other languages.

### DIFF
--- a/src/fuzz_introspector/utils.py
+++ b/src/fuzz_introspector/utils.py
@@ -382,7 +382,7 @@ def check_coverage_link_existence(link: str) -> bool:
 def _find_all_source_path(extension: str) -> Set[str]:
     """Search the $OUT/$SRC directory to find paths of all Java source files."""
     # Use set to avoid duplication
-    java_source_path_list = set()
+    source_path_list = set()
 
     # Retrieve $OUT and $SRC from environment variables
     out_dir = os.environ.get('OUT', None)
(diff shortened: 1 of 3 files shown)


src/fuzz_introspector/utils.py
```diff
@@ -441,3 +441,48 @@ def copy_java_source_files(required_class_list: List[str]):
...
+        if os.path.isfile(dst):
+            # Skip duplicate files
+            continue
+
+        shutil.copy(python_source_path, dst)
+        count += 1
+        copied_source_path_list.append(filename)
+
+    # Store a list of existing source file paths for reference
+    with open(os.path.join(constants.SAVED_SOURCE_FOLDER, 'index.json'),
```
DavidKorczynski, line 469: What is this index used for? I see you only added basename files to it, is this intentional? There is an existing `.json` that holds paths of all files: https://github.com/ossf/fuzz-introspector/blob/4efa8c9de479067f5683abb6fdf491529d609e5d/src/fuzz_introspector/constants.py#L26C1-L26C18

DavidKorczynski (changes requested):

src/fuzz_introspector/utils.py
arthurscchan: Yes. It is intentional. This logic is shared between python and JVM projects, which list all the python/java source files in the source-code directories for further source reference for the webapp. For python, it is basename only because all the python source files are copied directly to the source-code directory. But for Java projects, the source-code directory preserve the original Java package directories structure. In order to allow a unified API in the webapp to access file lists (with directory structure) in the source-code directory correctly, this index.json is needed.

DavidKorczynski (approved):

PR #1795, opened 2024-10-23
webapp: fix light entrypoints when cloud is used

DavidKorczynski:


DavidKorczynski:
CI failure is not due to this PR. Likely due to missing introspector reports as there were some missing introspector runs the last few days.

AdamKorcz (approved):

PR #1797, opened 2024-10-23
webapp: routes: include light-only searching for source code APIs

DavidKorczynski:


DavidKorczynski:
CI fails because of unrelated reason. This is good to go.

AdamKorcz (approved):

PR #1826, opened 2024-11-20
[Rust]: Fix calltree discovery logic

arthurscchan:
This PR fixes the calltree discovery logic to correctly interpret function calls initiated from fuzzing harness for rust frontend.

### DIFF
--- a/frontends/rust/rust_function_analyser/src/analyse.rs
+++ b/frontends/rust/rust_function_analyser/src/analyse.rs
@@ -16,8 +16,10 @@
 use serde::{Deserialize, Serialize};
 use std::collections::{HashMap, HashSet};
 use std::fs;
-use syn::{Expr, ExprBlock, FnArg, ImplItem, ImplItemFn, Item, ItemFn, ItemImpl, ReturnType, Stmt, Visibility};
-use syn::spanned::Spanned;
+use syn::{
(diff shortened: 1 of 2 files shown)


arthurscchan:
The newest logic successfully identified most of the call, with some exceptions that is known to be limitations. for example, in the call tree for parse_response_multspaces, it does not identify any function calls within the unsafe block, it seems that the syn approach failed to identify those blocked code. Will be working on finding alternative ways to identify those call.
```
    fn parse_with_config(&mut self, buf: &'b [u8], config: &ParserConfig) -> Result<usize> {
        let headers = core::mem::replace(&mut self.headers, &mut []);

        /* SAFETY: see `parse_headers_iter_uninit` guarantees */
        unsafe {
            let headers: *mut [Header<'_>] = headers;
            let headers = headers as *mut [MaybeUninit<Header<'_>>];
            match self.parse_with_config_and_uninit_headers(buf, config, &mut *headers) {
                Ok(Status::Complete(idx)) => Ok(Status::Complete(idx)),
                other => {
                    // put the original headers back
                    self.headers = &mut *(headers as *mut [Header<'_>]);
                    other
                },
            }
        }
    }
    ```

DavidKorczynski (approved):

PR #1843, opened 2024-11-27
webapp: Fix wrong path for local inspector report path

arthurscchan:
After the addition of light approach, the local $OUT/inspector directory has been moved to $OUT/inspector/inspector. This PR fixes the db building script to match the new directory structure for local FI build.


DavidKorczynski (changes requested):
Hmm, is this really the case? I don't see that. `inspector/inspector` can sometimes happen if project is build twice using introspector. This, however, shouldn't be the case following `rsync` in C/C++.

I believe you see this folder structure because of https://github.com/google/oss-fuzz/blob/92d61f8231c6075e0f9f555e1d2c7eb5dfecfa34/infra/base-images/base-builder/compile#L337

arthurscchan:
This PR is not needed any more after the fix in oss-fuzz compile infra script.

PR #1885, opened 2024-12-12
FI: Remove unnecessary package

arthurscchan:
This PR removes unnecessary package which requires libgraphviz-dev.


DavidKorczynski (approved):

DavidKorczynski:
I think we can remove `networkx` as well

PR #1888, opened 2024-12-13
core: frontend: avoid unnecessary foldering

DavidKorczynski:


arthurscchan:
This PR would break the oss-fuzz-integration, please also change the oss-fuzz-patch.diff in order to follow the new format.

DavidKorczynski:
holding off with this since @arthurscchan is working on some things that would be nice to land first.

PR #1993, opened 2025-01-17
analysis: Change analysis interface to allow passing in properties

arthurscchan:
This PR changes the analysis API to allow passing in additional properties from CLI for processing. This PR also initialise a new analysis for source file line function searching. The actual works for the analysis will be done in later PR.


DavidKorczynski (changes requested):
I think the "properties" is too broad. We should add these options on a case-by-case basis, e.g. just have a lot of `--enable-xx` if needed.

DavidKorczynski (approved):

PR #2009, opened 2025-01-21
pylint: Fix pylint error in frontend_cpp

arthurscchan:
This PR fixes pyline error in frontend_cpp

### DIFF
--- a/src/fuzz_introspector/frontends/frontend_cpp.py
+++ b/src/fuzz_introspector/frontends/frontend_cpp.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 #
 ################################################################################
+"""Tree-sitter frontend for cpp."""
 
 from typing import Any, Optional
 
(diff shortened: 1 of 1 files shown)


src/fuzz_introspector/frontends/frontend_cpp.py
```diff
@@ -563,6 +569,10 @@ def dump_module_logic(self,
                           harness_source: str = '',
                           dump_output=True):
         """Dumps the data for the module in full."""
+        _ = entry_function
+        _ = harness_name
```
DavidKorczynski, line 573: can we do some `# pylint disable` stuff instead of this?

DavidKorczynski (changes requested):

src/fuzz_introspector/frontends/frontend_cpp.py
arthurscchan: yes, changed.

DavidKorczynski (approved):

PR #2010, opened 2025-01-21
pylint: Fix pylint error for frontend-rust

arthurscchan:
This PR fixes the pylint error for frontend-rust.

### DIFF
--- a/src/fuzz_introspector/frontends/frontend_rust.py
+++ b/src/fuzz_introspector/frontends/frontend_rust.py
@@ -22,7 +22,7 @@ from tree_sitter import Language, Node
 import logging
 import yaml
 
-#pylint: disable=no-name-in-module
+# pylint: disable=no-name-in-module
 from fuzz_introspector.frontends import datatypes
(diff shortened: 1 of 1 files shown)


src/fuzz_introspector/frontends/frontend_rust.py
```diff
@@ -576,6 +582,8 @@ def dump_module_logic(self,
                           harness_source: str = '',
                           dump_output: bool = True):
         """Dumps the data for the module in full."""
+        _ = entry_function
```
DavidKorczynski, line 585: can we use `# pylint: disable=unused-argument` or similar for these cases?

DavidKorczynski (changes requested):

src/fuzz_introspector/frontends/frontend_rust.py
arthurscchan: Yes, changed.

DavidKorczynski (approved):

PR #2067, opened 2025-02-06
frontend: Allow returning frontend yaml directly

arthurscchan:
This PR fixes the frontend for other language, following the C++ approach to allow retrieving in-memory report yaml directly without the need to dump it to intermediate location if the full process is executed. It only generalises those functions to allow better reuse and refactoring later.


DavidKorczynski (approved):
Thanks, also take note of https://github.com/ossf/fuzz-introspector/pull/2062 to avoid regenerating a lot of data in case there are multiple harnesses

PR #2087, opened 2025-02-11
update requirements.txt

phi-go:
update PyYAML to add support for Cython 3.x and Python 3.13
and add missing requirements for the webapp (needed for launch_minor_oss_fuzz.sh and main.py)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,7 @@ cxxfilt==0.3.0
 lxml==4.9.1
 matplotlib==3.10.0
 numpy==2.1.0
-PyYAML==6.0
+PyYAML==6.0.2
 soupsieve==2.2.1
 yapf==0.32.0
 pylint==3.0.0
--- a/tools/web-fuzzing-introspection/requirements.txt
+++ b/tools/web-fuzzing-introspection/requirements.txt
@@ -5,7 +5,13 @@ Jinja2==3.1.5
 MarkupSafe==2.1.2
 Werkzeug==3.0.6
 requests==2.32.3
-pyyaml==6.0
+pyyaml==6.0.2
 orjson==3.10.0
 flask_smorest==0.44.0
 marshmallow==3.21.0
+apispec==6.8
+webargs==8.6
+blinker==1.9
+certifi==2025.1.31
+idna==3.10
+urllib3==2.3


DavidKorczynski (changes requested):
Could you post a reproducer for triggering the missing dependencies for the webapp? The current requirements work fine as far as I'm aware, but probably something with Python versioning?

phi-go:
Yes, this is because of the Python version. My system Python is 3.13, which is not compatible with some packages, so I used https://github.com/astral-sh/uv to manage the environment. Actually, I tried with 3.12 and got the package issues but it seems that aetheris is not compatible with 3.12 anyway, though the webapp worked without it.

However, using python 3.11 no changes are necessary, so I guess it is better to abandon this PR but update the docs to mention that only Python 3.11 is supported and maybe add something like the following as an example setup.

```bash
git clone https://github.com/ossf/fuzz-introspector.git
cd fuzz-introspector

# Set up environment
uv init --no-workspace --python 3.11 .
uv python pin
uv venv
source .venv/bin/activate

# Add and install dependencies
uv add -r requirements.txt

cd tools/web-fuzzing-introspection
uv add -r requirements.txt

# Test run
cd app/static/assets/db/
./launch_minor_oss_fuzz.sh
cd ../../../
python3 ./main.py
```

DavidKorczynski:
> Yes, this is because of the Python version. My system Python is 3.13, which is not compatible with some packages, so I used https://github.com/astral-sh/uv to manage the environment. Actually, I tried with 3.12 and got the package issues but it seems that aetheris is not compatible with 3.12 anyway, though the webapp worked without it.
> 
> However, using python 3.11 no changes are necessary, so I guess it is better to abandon this PR but update the docs to mention the supported Python versions and maybe add something like the following as an example setup for the webapp.
> 
> ```shell
> git clone https://github.com/ossf/fuzz-introspector.git
> cd fuzz-introspector
> 
> # Set up environment
> uv init --no-workspace --python 3.11 .
> uv python pin
> uv venv
> source .venv/bin/activate
> 
> # Add and install dependencies
> uv add -r requirements.txt
> 
> cd tools/web-fuzzing-introspection
> uv add -r requirements.txt
> 
> # Test run
> cd app/static/assets/db/
> ./launch_minor_oss_fuzz.sh
> cd ../../../
> python3 ./main.py
> ```

I agree

Let's close this PR and I'll update docs in places. I think I'll not add the `uv` part, primarily because I don't know much about it, so will just say we support Python 3.11

Issue #2054, opened 2025-02-04
Detect fuzzing issues by considering past results


phi-go:
Hello, as part of some research we analyzed fuzzer performance degradation by looking at the reasons why fuzzing coverage reduces for C/C++ projects in OSS-Fuzz. We found that there are several types of issues that are easier to detect by comparing to past reports.

**I would be happy to implement these metrics if you are interested.**

-  Detecting coverage drops would be a generic way to detect degradation, this is already discussed here: https://github.com/google/oss-fuzz/issues/11398. Here a threshold would need to be decided, maybe percentage or absolute number of lines.
- A common reason for large coverage drops is the vendoring of third-party library code, though, sometimes also project specific code. If you agree that library code should not be included in the coverage measurement, large changes should cause an alert and be ignored. See [grpc-httpjson-transcoding](https://storage.googleapis.com/oss-fuzz-introspector/grpc-httpjson-transcoding/inspector-report/20250202/fuzz_report.html) as an example, which is by itself a few hundred lines of code with close to 100% coverage but vendored 100k lines of library code.
- Compare the fuzz targets over time. It sometimes happens that a project starts to have a partial build failure that only stops one (or few) fuzz target from building, while not necessarily causing a build failure issue to be created for the project. For example this happened with curl: https://github.com/google/oss-fuzz/issues/11398#issuecomment-1867707444
- The number of corpus entries is normally quite stable. But due to the way coverage is collected can fluctuate and drop to a fraction of the real size: https://github.com/google/oss-fuzz/issues/12986 and https://github.com/google/oss-fuzz/issues/11935. So this could be detected by looking at past corpus sizes. Though, if I understand correctly the seed corpus is combined across fuzz targets? Alternatively, a expected number of corpus entries for covered code branches/lines could be decided. For example covering 10k lines with five corpus entries does not seem like effective fuzzing.

This is also related to diffing runs: https://github.com/ossf/fuzz-introspector/issues/734

I can also provide more examples if you want, just wanted to keep it short.

DavidKorczynski:
I like these ideas a lot and would be more than happy to review PRs.

Regarding third-party code, then my personal position is that any third-party code in your target is from a security standpoint the same as your own code, as longs as it's reachable/triggerable from untrusted input. So I think it's a bit more nuanced than just excluding third-party code.

In general I like the direction of these ideas and would happy to land them. I think these would require most changes to be done in the webapp rather than core, but am happy in either case to review and get PRs landed.

[...]

Issue #1338, opened 2023-12-22
web app: on project profile page add graphs for historical progression of coverage per-fuzzer


DavidKorczynski:
Ref: https://github.com/google/oss-fuzz/issues/11398#issuecomment-1867713967

phi-go:
I think we can close this issue.

evverx:
Agreed. I think it can be closed.

I can see spurious coverage drops affecting specific fuzz targets:

![Image](https://github.com/user-attachments/assets/0fcfd02b-5591-45be-8447-bdc137a7f1a8)

With those graphs it should be easier to figure out how stable the data is in the context of using it to file bugs reports automatically. It doesn't look like it's stable enough currently.

PR #2102, opened 2025-02-17
per fuzzer coverage information in project overview

phi-go:
Adds per fuzzer summary.json (based on `https://storage.googleapis.com/oss-fuzz-coverage/{0}/reports-by-target/{1}/{2}/linux/summary.json`) to the project timestamp data and generates per fuzzer coverage plots for the project overview, example below.

This PR is mainly to understand if an approach like this would be acceptable, though to implement some metrics in https://github.com/ossf/fuzz-introspector/issues/2054 it is probably needed to have a separate stage that takes a look at project data over time and not only in report generation. 

Still this PR [addresses an issue](https://github.com/ossf/fuzz-introspector/issues/1338) but it might take too much screen space (especially for large numbers of fuzzers), it might also require too much storage space, data ingress and runtime for data generation, and could have noticeable effects on webpage load time as well.

Screenshot_20250217_162814

### DIFF
--- a/tools/web-fuzzing-introspection/app/webapp/templates/project-profile.html
+++ b/tools/web-fuzzing-introspection/app/webapp/templates/project-profile.html
@@ -264,7 +264,7 @@ max_fuzzer_count = 0;
 
 {% for project_timestamp in project_statistics %}
   {% if project_timestamp.coverage_data != None %}
-    code_coverage_lines_x.push("{{project_timestamp.date}}");
+    code_coverage_lines_x.push('{{project_timestamp.date}}');
     code_coverage_lines_y.push({{project_timestamp.coverage_data.line_coverage.percent}});
     fuzzer_count_y.push({{project_timestamp.fuzzer_count}});
     max_fuzzer_count = Math.max(max_fuzzer_count, {{project_timestamp.fuzzer_count}});
@@ -277,11 +277,11 @@ max_fuzzer_count = 0;
 
     {% if project_timestamp.per_fuzzer_coverage_data != None %}
       {% for (fuzzer, cov_data) in project_timestamp.per_fuzzer_coverage_data.items() %}
-        if (fuzzers.get("{{fuzzer}}") === undefined) {
-          fuzzers.set("{{fuzzer}}", {x: [], y: []});
+        if (fuzzers.get('{{fuzzer}}') === undefined) {
+          fuzzers.set('{{fuzzer}}', {x: [], y: []});
         }
-        var vv = fuzzers.get("{{fuzzer}}");
-        vv.x.push("{{project_timestamp.date}}");
+        var vv = fuzzers.get('{{fuzzer}}');
+        vv.x.push('{{project_timestamp.date}}');
         vv.y.push({{cov_data.percent}});
       {%endfor%}
     {% endif %}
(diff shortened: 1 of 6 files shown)


phi-go:
I will fix the CI issues tomorrow.

Is there a command to run these CI tests locally?

DavidKorczynski (commented):
This looks good! Could you use single quotes instead of double quotes where possible?

We don't have helper scripts for the `CI` yet. In general, if you have typing and run `yapf -i` on the modules, then you should be good. We'd be happy to land some scripts that does this locally though, it's about automating https://github.com/ossf/fuzz-introspector/blob/main/.github/workflows/main.yml and https://github.com/ossf/fuzz-introspector/blob/main/.github/workflows/mypy.yml

phi-go:
> This looks good! Could you use single quotes instead of double quotes where possible?

Good to hear. Of course, will update.

> Once the CI is green then we can land this. We don't have helper scripts for the ci yet.

Ok, sounds good. I'll see if I get around to those scripts.

> Just to confirm, are you able to see the output of the failed CI runs?

Yes, I can see the details [here](https://github.com/ossf/fuzz-introspector/actions/runs/13374005490/job/37348936516?pr=2102) and [here](https://github.com/ossf/fuzz-introspector/actions/runs/13374005492/job/37348931343?pr=2102).

phi-go:
I fixed the formatting. However, two coverage reports that are needed for the webapp-api-test are not available, so this check does not succeed, how should we proceed?

https://storage.googleapis.com/oss-fuzz-coverage/abseil-cpp/reports/20250216/linux/summary.json
https://storage.googleapis.com/oss-fuzz-coverage/leveldb/reports/20250216/linux/summary.json

phi-go:
I reworked the project report a bit now there is a per fuzzer summary table, see below (with some example errors). This also should be more efficient when displaying the report and the summary is now available in all-project-current.json.

Screenshot_20250219_144309

DavidKorczynski:
We shouldn't fail if there is no coverage in this manner. Something like that following solves it:

```sh
$ git diff ./
diff --git a/tools/web-fuzzing-introspection/app/static/assets/db/web_db_creator_from_summary.py b/tools/web-fuzzing-introspection/app/static/assets/db/web_db_creator_from_summary.py
index 42f3470..4b98b52 100644
--- a/tools/web-fuzzing-introspection/app/static/assets/db/web_db_creator_from_summary.py
+++ b/tools/web-fuzzing-introspection/app/static/assets/db/web_db_creator_from_summary.py
@@ -319,7 +319,8 @@ def prepare_code_coverage_dict(
         project_language: str) -> Optional[Dict[str, Any]]:
     """Gets coverage URL and line coverage total of a project"""
     line_total_summary = extract_code_coverage_data(code_coverage_summary)
-
+    if not line_total_summary:
+        return None
     coverage_url = oss_fuzz.get_coverage_report_url(project_name,
                                                     date_str.replace('-', ''),
                                                     project_language)
@@ -457,7 +458,9 @@ def extract_local_project_data(project_name, oss_fuzz_path,
 
     code_coverage_data_dict = prepare_code_coverage_dict(
         code_coverage_summary, project_name, '', project_language)
-
+    if not code_coverage_data_dict:
+        return
+    
     if cov_fuzz_stats is not None:
         all_fuzzers = cov_fuzz_stats.split("\n")
         if all_fuzzers[-1] == '':
@@ -719,7 +722,9 @@ def extract_project_data(project_name, date_str, should_include_details,
 
     code_coverage_data_dict = prepare_code_coverage_dict(
         code_coverage_summary, project_name, date_str, project_language)
-
+    if not code_coverage_data_dict:
+        return
+    
     per_fuzzer_cov = {}
     if cov_fuzz_stats is not None:
         all_fuzzers = cov_fuzz_stats.split("\n")

```

phi-go:
Oof, I see, my bad. Thank you for taking a look!

Only the first return should be necessary, neither the local nor the normal run needed this check before and no logic was changed in that regard. I updated as such.

DavidKorczynski (approved):

tools/web-fuzzing-introspection/app/webapp/templates/project-profile.html
```diff
@@ -131,11 +130,57 @@ <h2>Project: <a href="{{project_url}}"> {{ project.name }}</a></h2>
...
+                          <code style="color: unset; white-space: nowrap;">
+                            {{ fuzzer_data['current'] }}% (avg: {{ fuzzer_data['avg'] }}%, max: {{ fuzzer_data['max'] }}%)
+                          </code>
+                        </td>
+                        <td>
+                          <a href="{{fuzzer_data['report_url']}}">{{ fuzzer_data['report_date'] }}</a>
+                        </td>
+                        <td>
+                          {% if fuzzer_data['got_lost'] %}
+                            Fuzzer no longer available!
```
evverx, line 166: Looking at
Screenshot 2025-03-04 at 19 10 00
I have to say that it's really cool. That being said that fuzz target was turned off on OSS-Fuzz on purpose so it would probably be great if it was possible to somehow indicate that it's known or something like that. It should also probably help when fuzz targets get renamed (though that use case is different and should probably be automated). I'm not sure.

phi-go: Yeah, I agree, would be great to add have that feature. This would require some way to specify expected or disabled harnesses which currently does not exist as far as I know. This would probably need to be supported by oss-fuzz, sadly it seems PRs there are much slower to be reviewed.

For now this only considers the past 30 days, so this fuzzer will be removed after a month.

Though, if you have another idea how to approach this, please let me know.

PR #2114, opened 2025-02-25
Include metadata file names in summary.json.

phi-go:
The 'fuzzerLogFile-<idx>-<id>.data' (.yaml) files can have a random id, add their names to make it easy to find them from the summary.json. Do the same for covreport which can have multiple files per fuzzer.

This change will make it much easier to implement analyses based on fuzzerLogFiles for the webapp.


DavidKorczynski (approved):

DavidKorczynski:
Looks good, thanks!

PR #2136, opened 2025-04-24
Add project charter document

GeauxJD:
Working on updating the project's standard OpenSSF setup documentation.


DavidKorczynski (commented):
Thanks @GeauxJD  -- just confirming if this is ready for review? It would probably make sense to update a bit on the mission of FI in the charter before landing this as FI has extended scope the last year.

PR #2183, opened 2025-06-07
fix: Branch blockers API extraction and processing

ch097711:
Previously, branch_blocker information was parsed from `summary.json`. However, a previous commit moved the branch_blocker information to `branch-blockers.json`, but the parsing logic was not updated accordingly. As shown in the figure, this API has no return value. 

image

Therefore, this PR fixes the parsing of branch_blocker data to now be extracted from [`branch-blockers.json`](https://storage.googleapis.com/oss-fuzz-introspector/tinyxml2/inspector-report/20250605/branch-blockers.json).

image

### DIFF
--- a/tools/web-fuzzing-introspection/app/static/assets/db/web_db_creator_from_summary.py
+++ b/tools/web-fuzzing-introspection/app/static/assets/db/web_db_creator_from_summary.py
@@ -440,7 +440,7 @@ def extract_local_project_data(project_name, oss_fuzz_path,
     branch_pairs = extract_and_refine_branch_blockers(branch_blockers,
                                                       project_name)
     # Dump things we dont want to accummulate.
-    save_branch_blockers(branch_pairs, project_name)
+    # save_branch_blockers(branch_pairs, project_name)
 
     try:
         project_repository = oss_fuzz.try_to_get_project_repository(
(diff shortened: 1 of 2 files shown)


DavidKorczynski (approved):
This looks good. I'm a bit unsure if we should land this because we're having some issues with storage size -- the branch blockers can be quite heavy and we need to watch out for taking up a lot of space.

ch097711:
I noticed a comment in the `extract_local_project_data` function that seems to prevent saving blocker information via `save_branch_blockers`:

```python
    # Dump things we dont want to accummulate.
    #save_branch_blockers(branch_pairs, project_name)
```

[https://github.com/ossf/fuzz-introspector/blob/b194efbc0ed81808ff3a43948777b79179dec214/tools/web-fuzzing-introspection/app/static/assets/db/web_db_creator_from_summary.py#L462](https://github.com/ossf/fuzz-introspector/blob/b194efbc0ed81808ff3a43948777b79179dec214/tools/web-fuzzing-introspection/app/static/assets/db/web_db_creator_from_summary.py#L462)

However, this line is not commented out in the `extract_project_data` function:
[https://github.com/ossf/fuzz-introspector/blob/b194efbc0ed81808ff3a43948777b79179dec214/tools/web-fuzzing-introspection/app/static/assets/db/web_db_creator_from_summary.py#L751](https://github.com/ossf/fuzz-introspector/blob/b194efbc0ed81808ff3a43948777b79179dec214/tools/web-fuzzing-introspection/app/static/assets/db/web_db_creator_from_summary.py#L751)

If there's still a concern about storage and a need to prevent saving this blocker information, maybe we could comment out all calls to `save_branch_blockers`. I could then add a note to the documentation for the `branch_blockers` API to clarify that.

Thanks!

ch097711:
> Yeah, there's still a concern. Could you do this though? Then I'll run this locally and do some analysis on the memory consumption -- if it looks good we can enable it

I've commented out the `save_branch_blockers` calls and updated the Swagger note for the `/api/branch-blockers` endpoint.

Thanks!

PR #2301, opened 2026-04-23
Adapt Fuzz Introspector to LLVM 22 API changes

Barrentkk:
Key changes:

1. Debug info migration (LLVM 19+ RemoveDIs):
   - Replaced DbgDeclareInst/DbgVariableIntrinsic with DbgVariableRecord
   - Added M.convertToNewDbgValues() to normalize debug info format
   - Recover argument names from debug info using new DbgVariableRecord API

2. getSourceLanguage() API change (LLVM 22):
   - Returns DISourceLanguageName instead of raw enum value
   - Extract numeric value via getName() before passing to dwarf::LanguageString()

3. getFirstNonPHIOrDbgOrLifetime() return type change (LLVM 20):
   - Now returns BasicBlock::iterator directly instead of Instruction*
   - Simplified getBBDebugInfo() implementation for LLVM 22

4. Pass registration updates (patch-llvm.sh):
   - Handle LLVM_ABI visibility macros in InitializePasses.h (LLVM 21+)
   - Account for initializeXRayInstrumentationPass rename (LLVM 21+)

5. Performance optimization:
   - Fast path skip DI scan when all arguments already have IR names
   - Early termination once all missing argument names are resolved

Updated build scripts and documentation to target LLVM 22.1.0.

### DIFF
--- a/build_all.sh
+++ b/build_all.sh
@@ -18,46 +18,40 @@ set -ex
 BASE=$PWD
 BUILD_BASE=$BASE/build
 
-if [ -d $BUILD_BASE ]; then
+if [ -d "$BUILD_BASE/llvm-project" ] && [ -d "$BUILD_BASE/llvm-build" ]; then
   echo "Reusing set up (LLVM Source). Updating the LLVM plugin"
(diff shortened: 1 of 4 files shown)


Barrentkk:
@DavidKorczynski Please review this change

